PR #14, opened 2022-07-15
Added git commit template

felixschlegel:
Added the git commit template also used in the other `swift-server/` repos

### DIFF
--- a/dev/git.commit.template
+++ b/dev/git.commit.template
@@ -1,0 +1,14 @@
+One line description of your change
+
+Motivation:
+
+Explain here the context, and why you're making that change.
+What is the problem you're trying to solve.
+
+Modifications:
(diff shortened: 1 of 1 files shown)


dev/git.commit.template
```diff
@@ -0,0 +1,14 @@
...
+Motivation:
+
+Explain here the context, and why you're making that change.
+What is the problem you're trying to solve.
+
+Modifications:
+
+Describe the modifications you've done.
+
+Result:
```
FranzBusch, line 12: Nit:
```suggestion
# Motivation:

Explain here the context, and why you're making that change.
What is the problem you're trying to solve.

# Modifications:

Describe the modifications you've done.

# Result:
```

Issue #2, opened 2022-06-24
Docker image for building package


felixschlegel:
Make sure that the package builds inside of Docker

Issue #3, opened 2022-06-24
Expose `librdkafka` to Swift Package


felixschlegel:
Expose the `rdkafka.h` headers to the Swift package using `module map` + a shim header as the location of the `rdkafka.h` file relies on the OS

PR #15, opened 2022-07-15
Created Docker image with example program

felixschlegel:
### Motivation:

This PR closes tickets #2 and #3.

We use a Docker image to make sure that wrapping the C Library also
works on Linux machines.

### Modifications:

* implemented the module map exposing the `librdkafka` C library to our
  Swift program
* created example program called `ErrorPrinter` that prints all error
  types exposed by Kafka
* set up Docker workflow
* changed `swift-tools-version` to `5.6` as there is no `Swift 5.7 Docker
  Image` available at the moment

### Result:

The user can print out all Kafka error names both on macOS and Linux.

### Try it out

```bash
docker-compose -f docker/docker-compose.yaml run swift-kafka-gsoc
```


docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
+ARG swift_version=5.6
+ARG ubuntu_version=bionic
+ARG base_image=swift:$swift_version-$ubuntu_version
+FROM --platform=linux/x86_64 $base_image
```
FranzBusch, line 4: Can we not fix the platform here. Let it take the host platform. Otherwise people with ARM machines always need to cross compile.
```suggestion
FROM $base_image
```

docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
+ARG swift_version=5.6
+ARG ubuntu_version=bionic
```
FranzBusch, line 2: We should use `focal` here
```suggestion
ARG ubuntu_version=focal
```

FranzBusch (changes requested):
Looks great already! Can you fix-up the few comments and let the formatter run please.

Sources/Crdkafka/shim.h
```diff
@@ -0,0 +1,8 @@
+#ifndef __RD_KAFKA_SHIM_H__
+#define __RD_KAFKA_SHIM_H__
+#ifdef __APPLE__
+#include "/usr/local/include/librdkafka/rdkafka.h"
+#else
+#include "/usr/include/librdkafka/rdkafka.h"
```
t089, line 6: shouldn't this be 
`<librdkafka/rdkafka.h>`
?
If you use an absolute path, it will not work on machines where homebrew is not in `/usr/local/`, eg, on Apple Silicon Macs homebrew defaults to `/opt/homebrew/include/librdkafka/rdkafka.h`

felixschlegel: That's a good point!

I have changed it to `#include <librdkafka/rdkafka.h>` for now but this got me running into more problems on my Apple Silicon machine:

When using `#include <librdkafka/rdkafka.h>`, the compiler looks through a list of system directories for this header file. Apparently, `/opt/homebrew` is not part of that list but could be added using the `-I` Swift compiler flag. However, I do not know where I would invoke that flag since we are using a module map.

Alternatively, we could use absolute paths in our `shim.h` after checking the OS and the target architecture as follows:

```swift
#ifndef __RD_KAFKA_SHIM_H__
#define __RD_KAFKA_SHIM_H__
#ifdef __APPLE__
#ifdef __x86_64__
#include "/usr/local/include/librdkafka/rdkafka.h"
#else
#include "/opt/homebrew/include/librdkafka/rdkafka.h"
#endif
#else
#include "/usr/include/librdkafka/rdkafka.h"
#endif
#endif
```

Furthermore, we would have to add a linker flag that adds our homebrew directory to library link search path in order to successfully compile on Apple Silicon:

```swift
.executableTarget(
    name: "ErrorPrinter",
    dependencies: ["Crdkafka"],
    linkerSettings: [
        .unsafeFlags(["-L/opt/homebrew/lib"])
    ]
),
```

FranzBusch: To get this unblocked, I think we should move forward with the #ifdef approach. If that compiles on both x86 and ARM based Macs that install it via brew that should be enough.
Once this package is further along we can revisit this.

Just as a side note, we can never use unsafe flags so that option can’t be used at all.

t089: I think the trick is to set the `PKG_CONFIG` env variable to include the homebrew dir, then SwiftPM (and Xcode) will be able to figure it out without additional flags in the package manifest. I need to check an example from my mac.

FranzBusch: Setting an env variable for builds seems rather unhandy. I would prefer to hardcode this for the brew installations on Mac in the beginning. This should get local development unblocked.
I want us to focus on the APIs for Kafka soon rather than spending too much time on this one.

FranzBusch: I just checked this a bit more. From what I understand using `systemLibrary` should do the magic. I don't think providing an env variable is needed @t089 . Once I am back at my work desk tomorrow I will investigate this some more.

t089: Maybe, this would be great! Last time I ran into this, I checked this code which lead me to the env variable `PKG_CONFIG_PATH`: https://github.com/apple/swift-package-manager/blob/57af70f2f9f5645514a9ce8e6279995c701333b0/Sources/PackageLoading/PkgConfig.swift

t089: So, I think you are right, best not to spend too much time on this now. But for completeness sake, for me (M1 Mac) with homebrew in `/opt/homebrew` the project builds and runs from this branch (`09a6331`) with this env variable configured (fish shell):

```fish
env PKG_CONFIG_PATH=/opt/homebrew/lib/pkgconfig:/opt/homebrew/share/pkgconfig:/usr/local/lib/pkgconfig:/usr/lib/pkgconfig:/opt/homebrew/opt/openssl@1.1/lib/pkgconfig swift build
```

I need to explicitly add the `openssl@1.1` path, because by default 
> openssl@1.1 is keg-only, which means it was not symlinked into /opt/homebrew,
because macOS provides LibreSSL.
> [...]
> For pkg-config to find openssl@1.1 you may need to set:
>  set -gx PKG_CONFIG_PATH "/opt/homebrew/opt/openssl@1.1/lib/pkgconfig"

(from `brew info openssl@1.1`)

I think this is how the `systemLibrary` machinery is supposed to work. You can set the env in a terminal and then open Xcode / code from the terminal with (`xed Package.swift`) to make it work within Xcode as well (make sure Xcode is not running already).  

You can verify that your `pkg-config` setup is correct by checking the output from

```fish
pkg-config --cflags --libs rdkafka
```

For me this gives for example

```text
-I/opt/homebrew/Cellar/librdkafka/1.9.1/include -I/opt/homebrew/Cellar/zstd/1.5.2/include -I/opt/homebrew/Cellar/lz4/1.9.3/include -I/opt/homebrew/Cellar/openssl@1.1/1.1.1q/include -L/opt/homebrew/Cellar/librdkafka/1.9.1/lib -lrdkafka
```

t089: So I think this is what SwiftPM does: It tries to find `brew`, then calls `pkg-config` from homebrew to get locations of `.pc` files with `pkg-config --variable pc_path pkg-config` and then will parse and process the `rdkafka.pc` file and its dependencies to find the correct compiler/linker flags. It then fails with `libcrypto` because it is missing in the search path (because "keg-only"), which can be fixed by explicitly adding it through the `PKG_CONFIG_PATH` variable.

FranzBusch: Thanks for investigating this in such a depth @t089!

I was also under the impression that this PR was doing everything right, but good catch with the `openssl` dependency! 

I would propose that we just add this to the `README` and move forward for now @felixschlegel

felixschlegel: Wow, thank you @t089! Your investigation really put us forward here!

For now, I will add a comment to the `README` stating that Apple Silicon users should `build` using

```bash
env PKG_CONFIG_PATH=/opt/homebrew/opt/openssl@1.1/lib/pkgconfig swift build
```

Thanks a lot!

felixschlegel: @fkiwitt can you please check on your Intel Mac if we also need to set the environment variable here?

t089: I've just tried this on my Intel iMac: you need to specify the `PKG_CONFIG_PATH=/opt/homebrew/opt/openssl@1.1/lib/pkgconfig` on intel as well, otherwise the build will fail.

t089: I would recommend to add this to the README:

```zsh
PKG_CONFIG_PATH=$(brew --prefix)/opt/openssl@1.1/lib/pkgconfig swift build
```

This will work no matter where homebrew is installed.

Or to open the project in Xcode


```zsh
PKG_CONFIG_PATH=$(brew --prefix)/opt/openssl@1.1/lib/pkgconfig xed Package.swift
```

FranzBusch (approved):
Thanks for the work @felixschlegel and thanks @t089 for the investigation efforts!

felixschlegel:
Thank you @FranzBusch @t089!

I have learned a lot during this conversation! 🏫

PR #18, opened 2022-08-15
Basic implementation of the `KafkaClient` class

### LINKED ISSUE #5: Enable connection to Kafka Servers
> **Note**: Use the [GSoC22 Proposal](https://github.com/felixschlegel/ProposalSwiftGSoC22/blob/main/main.pdf) as a reference for the implementation of the public interface.

* make it possible to **connect** and **disconnect** to/from a Kafka Server by implementing the `KafkaClient` class
* create a static `Logger` property for logging the client status

    **Not in the scope** of this issue:

* customizable `KafkaConfig`, please use default config here
* `connectAdditional()` method, just throw a `fatalError()` or leave it out completely

felixschlegel:
This PR closes #5 .

## Motivation

The `KafkaClient` class is the base class for the future
implementations of `KafkaProducer` and `KafkaConsumer`. It handles the
connection to the Kafka C library.

## Modifications

* removed the `ErrorPrinter` example code
* added a basic implementation for `KafkaConfig`
* added a basic implementation for `KafkaError`
* deleted boilerplate `SwiftKafka` `struct`
* implemented `KafkaClient`
* added `swift-log` to project
* added DocC documentation to `KafkaClient`

## Result

There is a `KafkaClient` class that is configurable using a preliminary
implementation of `KafkaConfig`.

## Additional Information

Initially, I planned to write tests that check if the `KafkaClient` is able to connect to a Kafka Cluster. However, I my tests would always succeed even though I was passing invalid bootstrap server domains to the `KafkaClient`. This might be related to [this issue](https://github.com/edenhill/librdkafka/issues/2062).


Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -0,0 +1,67 @@
...
+
+            let configResult = rd_kafka_conf_set(
+                pointer,
+                key,
+                newValue,
+                errorString,
+                KafkaClient.stringSize
+            )
+
+            // TODO: what shall we do when the value could not be set?
```
FranzBusch, line 64: Can we make this a throwing set?

Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -0,0 +1,67 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+
+public class KafkaConfig {
```
FranzBusch, line 17: Can we add a doc comment here please. Also pointing to the librdkafka docs.

I am inclined to call this `LIBRDKafkaConfig`

Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    public init() {
+        self.pointer = rd_kafka_conf_new()
+    }
+
+    deinit {
+        rd_kafka_conf_destroy(pointer)
+    }
+
+    public subscript(key: String) -> String? {
```
FranzBusch, line 33: Can we add some docs here

Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -0,0 +1,67 @@
...
+
+public class KafkaConfig {
+    // Preliminary implementation
+    private var pointer: OpaquePointer
+
+    func getPointerDuplicate() -> OpaquePointer {
+        return rd_kafka_conf_dup(self.pointer)
+    }
+
+    public init() {
```
FranzBusch, line 25: Can we add some docs here

Sources/SwiftKafka/KafkaClient.swift
```diff
@@ -0,0 +1,68 @@
...
+/// Base class for ``KafkaProducer`` and ``KafkaConsumer``, which is used to handle the connection to the Kafka ecosystem.
+public class KafkaClient {
+    // Default size for Strings returned from C API
+    static let stringSize = 1024
+
+    private var _kafkaHandle: OpaquePointer?
+    private var _clientType: rd_kafka_type_t
+    private var _config: KafkaConfig
+
+    static let logger = Logger(label: "SwiftKafkaLogger")
```
FranzBusch, line 27: We should get the logger passed in and not instantiate one of our own. Make sure to give this guide a read: https://www.swift.org/server/guides/libraries/log-levels.html

FranzBusch (changes requested):
Off to a great start! I left some comments inline

Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -0,0 +1,67 @@
...
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+
+public class KafkaConfig {
+    // Preliminary implementation
+    private var pointer: OpaquePointer
+
+    func getPointerDuplicate() -> OpaquePointer {
+        return rd_kafka_conf_dup(self.pointer)
```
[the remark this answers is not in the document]
felixschlegel: Good catch! How do you like this fix? [d5fcccd](https://github.com/swift-server/swift-kafka-gsoc/pull/18/commits/d5fcccda56020a68e7f0146cc3f42ab02dcf2d7c)

Sources/SwiftKafka/KafkaConfig.swift
felixschlegel: Unfortunately, that is not possible :(

`'set' accessor cannot have specifier 'throws'`

felixschlegel: We could have explicit functions like `set(value:, forKey:)` and `value(forKey:)` here — maybe it's best to postpone the discussion about `KafkaConfig`'s public interface to its own dedicated PR and try to get things running first

Sources/SwiftKafka/KafkaConfig.swift
felixschlegel: IMO, it depends on how we want to expose the config Kafka config to the user — I would stick with the `KafkaConfig` name for the public interface. However, if we want to abstract these key/value pairs e.g. by having a struct with strongly typed configuration properties, we can consider having an `internal` `LIBRDKafka` class for the key/value pair implementation that is then accessed by our `public` `KafkaConfig` struct.

Sources/SwiftKafka/KafkaClient.swift
```diff
@@ -0,0 +1,61 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+import Logging
+
+/// Base class for ``KafkaProducer`` and ``KafkaConsumer``, which is used to handle the connection to the Kafka ecosystem.
+class KafkaClient {
```
FranzBusch, line 19:
```suggestion
final class KafkaClient {
```

Sources/SwiftKafka/KafkaClient.swift
```diff
@@ -0,0 +1,61 @@
...
+    // Default size for Strings returned from C API
+    static let stringSize = 1024
+
+    let logger: Logger
```
FranzBusch, line 23:
```suggestion
    /// A logger.
   private let logger: Logger
```

Sources/SwiftKafka/KafkaClient.swift
```diff
@@ -0,0 +1,61 @@
...
+
+    private let _clientType: rd_kafka_type_t
+    private let _config: KafkaConfig
+    private let _kafkaHandle: OpaquePointer
```
FranzBusch, line 27: I think we can drop the `_` here and could we add some doc comments what each of these are.

Sources/SwiftKafka/KafkaConfig.swift
FranzBusch: I think we need to go the route of having a non-strongly typed struct in the beginning. This leaves the greatest flexibility. 

Furthermore, I think we should slightly change this type. Instead of it being a class we should make it a struct with an internal class. The struct should implement copy-on-write semantics. In the case of a copy we would dupe the pointer. This allows us to not dupe the pointer when we pass it to librdkafka. You probably want to use this: https://developer.apple.com/documentation/swift/isknownuniquelyreferenced(_:)-98zpp

FranzBusch (commented):
Had some more time to think about this and left some comments inline

felixschlegel:
I have implemented a copy-on-write mechanism for `KafkaConfig` and included some tests, let me know what you think!

Sources/SwiftKafka/KafkaClient.swift
```diff
@@ -0,0 +1,69 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+import Logging
+
+/**
+ Base class for ``KafkaProducer`` and ``KafkaConsumer``,
+ which is used to handle the connection to the Kafka ecosystem.
+ */
```
FranzBusch, line 21: This is a super nit but can we use: `///` please instead of `/**`. This aligns with the style we use in almost all Swift on Server repos

Sources/SwiftKafka/KafkaClient.swift
FranzBusch: This is still open

Sources/SwiftKafka/KafkaClient.swift
```diff
@@ -0,0 +1,69 @@
...
+        self.kafkaHandle = handle
+    }
+
+    deinit {
+        rd_kafka_destroy(kafkaHandle)
+    }
+
+    func connectAdditional(brokers: [String]) {
+        fatalError("Not implemented")
+    }
```
FranzBusch, line 68: Can we remove this for now since we don't use it anywhere anyhow?

Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -0,0 +1,112 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+
+/**
+ Class that is used to configure producers and consumers.
+ Please see the [list of all available configuration properties](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md) for more information.
+ */
```
FranzBusch, line 20: Same here

FranzBusch: Can we also add here that this is a CoW structure?

Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -0,0 +1,112 @@
...
+
+    private var _internal: _Internal
+
+    public init() {
+        self._internal = .init()
+    }
+
+    public var pointer: OpaquePointer {
+        return self._internal.pointer
+    }
```
FranzBusch, line 90: I don't think this should be public

Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -0,0 +1,112 @@
...
+
+        try self._internal.set(value, forKey: key)
+    }
+}
+
+extension KafkaConfig: Equatable {
+    public static func == (lhs: KafkaConfig, rhs: KafkaConfig) -> Bool {
+        return lhs._internal === rhs._internal
+    }
+}
```
FranzBusch, line 112: Why isn't the default `Equatable` conformance working here? `OpaquePointer` conforms to `Equatable`. 

Also if we had `Equatable` we should probably just add `Hashable` as well.

FranzBusch (commented):
Very close to merging this. Some last comments

Sources/SwiftKafka/KafkaClient.swift
felixschlegel: My bad, have not noticed that it should be `private`

Sources/SwiftKafka/KafkaConfig.swift
felixschlegel: According to the build logs, `_Internal` should conform to `Equatable` if we want to use the default conformance. But then we would have to somehow make `_Internal` conform to `Equatable` as well — for me the default conformance did not work here either

FranzBusch: Can you please check again. This works:
```swift
struct Foo: Hashable {
    let pointer: OpaquePointer
}
```

Also can you conform to `Hashable` instead of `Equatable`

PR #20, opened 2022-08-19
Auxiliary types for sending Messages to Kafka

felixschlegel:
## Motivation:

To be able to send messages to Kafka we need auxiliary types such as

* `KafkaProducerMessage`
* `KafkaTopicConfig`

## Modifications:

* implemented `KafkaProducerMessage`
* created implementation for `KafkaTopicConfig` that matches
  `KafkaConfig`s copy-on-write mechanism and has tests
* created barebones of `KafkaProducer` class

### DIFF
--- a/Sources/SwiftKafka/KafkaProducer.swift
+++ b/Sources/SwiftKafka/KafkaProducer.swift
@@ -15,9 +15,12 @@
 import Crdkafka
 import Logging
 
+/// Send messages to the Kafka cluster.
+/// - Note: When messages get published to a non-existent topic, a new topic is created using the ``KafkaTopicConfig``
+/// configuration object (only works if server has `auto.create.topics.enable` property set).
 public struct KafkaProducer {
-    let client: KafkaClient
-    let topicConfig: KafkaTopicConfig
+    private let client: KafkaClient
+    private let topicConfig: KafkaTopicConfig
 
     // Preliminary implementation
     public init(
--- a/Sources/SwiftKafka/KafkaProducer.swift
+++ b/Sources/SwiftKafka/KafkaProducer.swift
@@ -1,0 +1,31 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the swift-kafka-gsoc open source project
+//
+// Copyright (c) 2022 Apple Inc. and the swift-kafka-gsoc project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+import Logging
+
+public struct KafkaProducer {
+    let client: KafkaClient
+    let topicConfig: KafkaTopicConfig
+
+    // Preliminary implementation
+    public init(
+        config: KafkaConfig = KafkaConfig(),
+        topicConfig: KafkaTopicConfig = KafkaTopicConfig(),
+        logger: Logger
+    ) throws {
+        self.client = try KafkaClient(type: .producer, config: config, logger: logger)
+        self.topicConfig = topicConfig
+    }
+}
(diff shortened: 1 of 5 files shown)


Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -0,0 +1,31 @@
...
+// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+import Logging
+
+public struct KafkaProducer {
```
FranzBusch, line 18: Let's add some doc comments here

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -0,0 +1,31 @@
...
+    let client: KafkaClient
+    let topicConfig: KafkaTopicConfig
```
FranzBusch, line 20: Can these be private?

Sources/SwiftKafka/KafkaProducer.swift
FranzBusch: We will see but I think this will be a `final class` in the long run

Sources/SwiftKafka/KafkaProducerMessage.swift
```diff
@@ -0,0 +1,65 @@
...
+
+import Crdkafka
+import Foundation
+
+/// Message that is sent by the `KafkaProducer`
+public struct KafkaProducerMessage {
+    let topic: String
+    let partition: Int32
+    let key: ContiguousBytes?
+    let value: ContiguousBytes
```
FranzBusch, line 23: Can we make these `public var`s? This would allow the user to change stuff after init which is always handy.

Sources/SwiftKafka/KafkaProducerMessage.swift
```diff
@@ -0,0 +1,65 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+import Foundation
+
+/// Message that is sent by the `KafkaProducer`
+public struct KafkaProducerMessage {
+    let topic: String
+    let partition: Int32
```
FranzBusch, line 21: Maybe it is useful to create a concrete type for partition that can be created with a `Int32` value and has a `static let` that is the `RD_KAFKA_PARTITION_UA` value under the hood.

Sources/SwiftKafka/KafkaTopicConfig.swift
```diff
@@ -0,0 +1,121 @@
...
+// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+
+/// When messages get published to a non-existent topic, a new topic is created using the `KafkaTopicConfig`
+/// configuration object (only works if server has `auto.create.topics.enable` property set).
```
FranzBusch, line 18: This seems like a very specific first sentence for this type. Maybe we can start with the more general docs here first and add this as a `- Note:` block

FranzBusch (changes requested):
Looks very good already! Left some inline comments.

Sources/SwiftKafka/KafkaProducer.swift
felixschlegel: > Let's add some doc comments here

Implemented in PR on top of this one https://github.com/felixschlegel/swift-kafka-gsoc/pull/1

Sources/SwiftKafka/KafkaTopicConfig.swift
felixschlegel: Good point. I have moved it to the documentation of `KafkaProducer` as it makes more sense for the user (IMO).

FranzBusch (commented):
One small thing

FranzBusch (approved):

PR #21, opened 2022-08-23
`KafkaProducer` with `sendAsync` method

felixschlegel:
## Motivation

Enable users to send messages in a non-blocking way with an optional callback.

## Modifications

* specified minimum OS versions in `Package.swift` to support new Swift
  concurrency
* made it possible to test the KafkaProducer inside of Docker
* `KafkaProducerMessage`
* `KafkaConsumerMessage`
* `KafkaProducer `with sendAsync method
* initialization of `KafkaError `from `rd_kafka_resp_err_t`


Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -25,18 +25,20 @@ public struct KafkaConfig: Hashable, Equatable {
...
+        }
+
+        /// Initialize internal `KafkaConfig` object with default configuration
+        convenience init() {
+            self.init(pointer: rd_kafka_conf_new())
         }
 
         deinit {
-            rd_kafka_conf_destroy(pointer)
+            // rd_kafka_conf_destroy(pointer)
```
felixschlegel, line 40: We are running into an issue here:

The `pointer` to the `rd_kafka_conf_t` gets automatically freed by `rd_kafka_new`. This leads us to a situation where we have a double-freeing of this pointer. However, when the `KafkaConfig` is created without passing it to `rd_kafka_new`, we still need to deallocate it here.

A solution could be checking if the pointer has been freed or not. Though I don't know whether this is possible.

Sources/SwiftKafka/KafkaConsumerMessage.swift
```diff
@@ -0,0 +1,78 @@
...
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+import struct Foundation.Data
+
+/// Messages that are received from the Kafka cluster.
+public struct KafkaConsumerMessage {
+    let topic: String
+    let partition: Int32
+    let key: Data?
```
felixschlegel, line 22: According to [the proposal](https://github.com/felixschlegel/ProposalSwiftGSoC22/blob/main/main.pdf), we wanted `key` and `value` to be `ContiguousBytes`. However, we are just reading this `struct` and not modifying anything. Wouldn't it be more convenient to go with a concrete implementation of `ContiguousBytes` here such as `Data`?

Sources/SwiftKafka/KafkaConsumerMessage.swift
```diff
@@ -0,0 +1,78 @@
...
+    let value: Data
+    let offset: Int64
+
+    // TODO: copy on write, does this even make sense as we read-only?
```
felixschlegel, line 26: We agreed on also having a copy-on-write mechanism for `KafkaConsumerMessage`. As Kafka messages are just read, why shouldn't we keep it as a read-only `struct` or have a `struct` that references an `_Internal` class that contains all the data?

Sources/SwiftKafka/KafkaTopicConfig.swift
```diff
@@ -0,0 +1,122 @@
...
+            self.pointer = rd_kafka_topic_conf_new()
+        }
+
+        /// Initialize internal `KafkaTopicConfig` object through a given `rd_kafka_topic_conf_t` pointer
+        init(pointer: OpaquePointer) {
+            self.pointer = pointer
+        }
+
+        deinit {
+            // rd_kafka_topic_conf_destroy(pointer)
```
felixschlegel, line 40: We are running into an issue here:

The `pointer` to the `rd_kafka_topic_conf_t` gets automatically freed by `rd_kafka_topic_new`. This leads us to a situation where we have a double-freeing of this pointer. However, when the `KafkaTopicConfig` is created without passing it to `rd_kafka_topic_new`, we still need to deallocate it here.

A solution could be checking if the pointer has been freed or not. Though I don't know whether this is possible.

Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -80,6 +82,41 @@ public struct KafkaConfig: Hashable, Equatable {
...
 
+        /// Message callback has to be instantiated before KafkaClient is initialized
+        /// as pointer to rd_kafka_conf_t gets freed after rd_kafka_new()
+        func setMessageCallback() {
+            rd_kafka_conf_set_dr_msg_cb(
+                self.pointer,
+                { kafkaHandle, message, _ in
+
+                    guard let kafkaHandle = kafkaHandle, let message = message else {
+                        return // TODO: log error
```
felixschlegel, line 93: I would like to log an error here, however, we cannot capture values in C closures. Do you have any idea how we could access the `logger` of the `KafkaProducer` from this closure?

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -13,22 +13,183 @@
...
 import Logging
 
 /// Send messages to the Kafka cluster.
 /// - Note: When messages get published to a non-existent topic, a new topic is created using the ``KafkaTopicConfig``
 /// configuration object (only works if server has `auto.create.topics.enable` property set).
-public struct KafkaProducer {
+public actor KafkaProducer {
+    /// Class that lets us wrap both the `messageID` and the `producer` into a single pointer
+    private final class CallbackOpaquePointer {
+        let messageID: UUID // TODO: to avoid Foundation, have a counter in KafkaProducer that serves as ID
```
felixschlegel, line 25: I went with `UUID` here, but if we want to stay away from `Foundation` I could imagine having an isolated `counter` variable that can get passed as a message-id.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -13,22 +13,183 @@
...
+                keyBytes?.count ?? 0,
+                pointer
+            )
+        }
+
+        guard responseCode == 0 else {
+            throw KafkaError(rawValue: responseCode)
+        }
+
+        // TODO: can messages be acked before the message gets set? Should not be possible -> we are an actor
```
felixschlegel, line 137: Just to double-check, the question should rather be

`an messages be acked before the message callback gets set?`

However, I think the `actor` keeps us safe here

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -13,22 +13,183 @@
...
+        self.receivedMessages.append(message)
+    }
+
+    /// `AsyncStream` that returns all ``KafkaProducerMessage`` objects that have been
+    /// acknowledged by the Kafka cluster.
+    public nonisolated var acknowledgements: AsyncStream<KafkaProducerMessage> {
+        AsyncStream { continuation in
+            Task {
+                while !Task.isCancelled {
+                    // TODO: Sleep here to slow down rate?
```
felixschlegel, line 156: Should we reduce the rate to something like every millisecond?

Tests/SwiftKafkaTests/KafkaProducerTests.swift
```diff
@@ -0,0 +1,98 @@
...
+        let expectedTopic = "test-topic"
+        let message = KafkaProducerMessage(
+            topic: expectedTopic,
+            value: "Hello, World!"
+        )
+
+        try await producer.sendAsync(message: message)
+
+        for await acknowledgedMessage in producer.acknowledgements {
+            // TODO: make Producer Message Equatable somehow and test entire message -> Problem: Hard to make Contiguous bytes equatable
```
felixschlegel, line 55: I know `ContiuousBytes` gives the user a lot of flexibility — however, we often run into pain points e.g. we cannot make it `Equatable` which makes testing hard

FranzBusch (commented):
I have been thinking about this some more and I think we still need to do it slightly different. I think we should get rid of the `CallbackOpaquePointer` again and actually use the opaque pointer on the topic to pass a reference to `self`. We need to unretain this passed pointer in the `shutdown` method after we called `rd_kafka_destroy`.
For the id of messages we can do a neat little trick we can use a counter on the actor of `UInt`s and pass this to the `rd_kafka_produce`:
`let pointer = UnsafeMutableRawPointer(bitPattern: UInt(id))`

I also left some more comments inline

Sources/SwiftKafka/KafkaConsumerMessage.swift
```diff
@@ -0,0 +1,78 @@
...
+    // TODO: deallocate rd_kafka_message_t on deinit
+
+    /// Initialize `KafkaConsumerMessage` from `rd_kafka_message_t` pointer.
+    init(messagePointer: UnsafePointer<rd_kafka_message_t>) throws {
+        let rdKafkaMessage = messagePointer.pointee
```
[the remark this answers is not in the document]
felixschlegel: We definitely need to call `rd_kafka_message_destroy`, thanks for pointing that out! However, I would prefer calling it in the `KafkaProducer:deliveryReportCallback` closure as this one actually owns the pointer to the `rd_kafka_message_t`

Sources/SwiftKafka/KafkaConsumerMessage.swift
```diff
@@ -0,0 +1,78 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+import struct Foundation.Data
+
+/// Messages that are received from the Kafka cluster.
+public struct KafkaConsumerMessage {
```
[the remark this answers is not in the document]
felixschlegel: Ok, I will implement `KafkaConsumerMessage` in the `KafkaConsumer` PR then.

In this PR I have now named it `KafkaAckedMessage`.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -13,22 +13,188 @@
...
+
+        for (_, topicHandle) in self.topicHandles {
+            rd_kafka_topic_destroy(topicHandle)
+        }
+
+        callbackOpaquePointers.forEach {
+            _ = Unmanaged<CallbackOpaquePointer>.fromOpaque($0).takeRetainedValue()
+        }
+
+        self.pollTask.cancel()
```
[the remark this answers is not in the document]
felixschlegel: So technically, the `deinit` of `KafkaProducer` gets called now so `rd_kafka_destroy` would be called automatically by `KafkaClient`s `deinit`. However, we have to figure out a way on what to do it when the user tries to send a message from a `KafkaProducer` that has already been shut down.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,179 @@
...
+        let messageID = messagePointer.pointee._private
+        guard let message = KafkaAckedMessage(messagePointer: messagePointer, id: UInt(bitPattern: messageID)) else {
+            Task {
+                producer.logger.error("Sending message failed due to an unknown error")
+            }
+            return
+        }
+
+        // TODO: is this thread-safe???
+        producer.acknowledgements.produceMessage(message)
```
felixschlegel, line 172: Do you think this is safe to do?

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,179 @@
...
     private let topicConfig: KafkaTopicConfig
+    private let logger: Logger
+    private var topicHandles: [String: OpaquePointer]
 
-    // Preliminary implementation
+    // Implicitly unwrapped optional because we must pass self to our config before initializing client etc.
+    private var referenceToSelf: UnsafeMutableRawPointer!
+    private var client: KafkaClient!
+    private var kafkaHandle: OpaquePointer!
+    private var pollTask: Task<Void, Never>!
```
felixschlegel, line 70: Let me know what you think about these implicitly unwrapped optionals 😅

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,183 @@
...
+        self.config.setCallbackOpaque(opaque: self.referenceToSelf)
+
+        self.client = try KafkaClient(type: .producer, config: self.config, logger: self.logger)
+        self.kafkaHandle = self.client.kafkaHandle
+
+        // Poll Kafka every millisecond
+        self.pollTask = Task { [kafkaHandle] in
+            while !Task.isCancelled {
+                rd_kafka_poll(kafkaHandle, 0)
+                try? await Task.sleep(nanoseconds: 1_000_000)
```
FranzBusch, line 97: Can we use the `Clock` API here instead of the nanosecond one?

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,183 @@
...
+            }
+        }
+    }
+
+    /// Kafka producer must be closed **explicitly**.
```
FranzBusch, line 102:
```suggestion
    /// Method to shutdown the ``KafkaProducer``.
    ///
    /// This method flushes any buffered messages and waits until a callback is received for all of them.
    /// Afterwards, it shuts down the connection to Kafka and cleans any remaining state up.
```

Sources/SwiftKafka/KafkaProducer.swift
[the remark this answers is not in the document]
FranzBusch: This is a good point. I think we have to to introduce a little state keeping here. A small state enum like this should do the trick
```swift
enum State {
    case started
    case shuttingDown
    case shutdown
}
```

In the shutdown method you to switch over the current state and depending on that decide what to do and transition to the correct state.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,183 @@
...
+    /// This function is non-blocking.
+    /// - Parameter message: The ``KafkaProducerMessage`` that is sent to the KafkaCluster.
+    /// - Returns: Unique message identifier matching the `id` property of the corresponding ``KafkaAckedMessage``
+    @discardableResult
+    public func sendAsync(message: KafkaProducerMessage) throws -> UInt {
+        let topicHandle = self.createTopicHandleIfNeeded(topic: message.topic)
+
+        let keyBytes: [UInt8]?
+        if let key = message.key {
+            keyBytes = key.getBytes(at: 0, length: key.readableBytes)
```
FranzBusch, line 125: Using the `getX` methods on `ByteBuffer` is almost never correct since it does not respect the current `readerIndex` and doesn't move it forward. Please use the `readBytes` method instead.  This will require you to create a local copy of the message which is okay. In the future, we could mark the `message` parameter as `taking` to indicate that it takes ownership of the value

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,183 @@
...
+        }
+
+        self.messageIDCounter += 1
+
+        let responseCode = message.value.withUnsafeReadableBytes { valueBuffer in
+
+            return rd_kafka_produce(
+                topicHandle,
+                message.partition.rawValue,
+                RD_KAFKA_MSG_F_COPY,
```
FranzBusch, line 137: Can we document what this does

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,183 @@
...
+        }
+
+        return self.messageIDCounter
+    }
+
+    // Closure that is executed when a message has been acknowledged by Kafka
+    private let deliveryReportCallback: (
+        @convention(c) (OpaquePointer?, UnsafePointer<rd_kafka_message_t>?, UnsafeMutableRawPointer?) -> Void
+    ) = { _, messagePointer, producerPointer in
+
```
FranzBusch, line 157: I would prefer if we try to unwrap the `opaquePointer` (here called `producerPointer`) first. If it is `nil` we should `fatalError` because something in our code went wrong.

Once, we have the reference to `self` we should call a method on self that passes the `messagePointer`. This way we get off the C callback as fast as possible.

Inside this new method we should add least add a debug log when we get a callback for message and the `messagePointer` is `nil`.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,183 @@
...
+        guard let messagePointer = messagePointer else {
+            return
+        }
+
+        guard let producerPointer = producerPointer else {
+            return
+        }
+        let producer = Unmanaged<KafkaProducer>.fromOpaque(producerPointer).takeUnretainedValue()
```
FranzBusch, line 165: Can we add a comment here why we are taking unretained and how/where we balance it.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,183 @@
...
         logger: Logger
-    ) throws {
-        self.client = try KafkaClient(type: .producer, config: config, logger: logger)
+    ) async throws {
+        self.config = config
         self.topicConfig = topicConfig
+        self.logger = logger
+        self.topicHandles = [:]
+
+        self.referenceToSelf = Unmanaged.passRetained(self).toOpaque()
```
FranzBusch, line 86: Can we add a comment to this why we `passRetained` and where we balance it.

Sources/SwiftKafka/KafkaProducer.swift
FranzBusch: Yes this is thread safe because the underlying `NIOAsyncSequenceProducer` is thread safe.

Sources/SwiftKafka/KafkaProducer.swift
FranzBusch: We also need to call [rd_kafka_message_destroy](https://docs.confluent.io/2.0.0/clients/librdkafka/rdkafka_8h.html#a35e0c519209d1afe9e29468f766c1c24) in this new method

Tests/SwiftKafkaTests/KafkaProducerTests.swift
```diff
@@ -0,0 +1,102 @@
...
+// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@testable import SwiftKafka
+import XCTest
+
+final class KafkaProducerTests: XCTestCase {
```
FranzBusch, line 18: Can we add an assert that after calling `shutdownGracefully` that the `KafkaProducer` is deinited. You can do this by creating a weak ref before then shutting down and dropping the strong ref. Afterwards, you can assert that the weak ref is `nil`.

FranzBusch (commented):
I think we are getting really close here. Some more comments

Sources/SwiftKafka/KafkaProducer.swift
felixschlegel: (After chatting, we agreed on waiting until `Swift 5.7` is officially out)

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,246 @@
...
+struct NoDelegate: NIOAsyncSequenceProducerDelegate {
+    func produceMore() {}
+    func didTerminate() {}
+}
+
+/// `AsyncSequence` implementation for handling messages acknowledged by the Kafka cluster (``KafkaAcknowledgedMessage``).
+public struct AcknowledgedMessagesAsyncSequence: AsyncSequence {
+    public typealias Element = KafkaAcknowledgedMessage
+
+    let _internal = NIOAsyncSequenceProducer.makeSequence(
```
FranzBusch, line 35: We should not be holding the result of the `makeSequence` call here. This way you are retaining both the `NIOAsyncSequenceProducer` and the `NIOAsyncSequenceProducer.Source`. We should rather pass in the former into an init of this sequence and hold the `Source` inside the `KafaProducer`. This means calling `makeSequence` in the `init` of the producer.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,246 @@
...
+        }
+    }
+
+    /// Method to shutdown the ``KafkaProducer``.
+    ///
+    /// This method flushes any buffered messages and waits until a callback is received for all of them.
+    /// Afterwards, it shuts down the connection to Kafka and cleans any remaining state up.
+    public func shutdownGracefully() throws {
+        try failIfProducerShutDown()
+        self.state = .shuttingDown
```
FranzBusch, line 146: Instead of doing this can you please explicitly `switch` over the state here. This is a common pattern in the swift on server space to make sure we are only doing state transition at the right time.

I would recommend to have a `switch` over the state in this method and then a separate `private` method called `_ shutdownGracefully` that does the actual shutdown and only call this in the case of `.started`

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,246 @@
...
+        self.state = .shutDown
+    }
+
+    /// Send messages to the Kafka cluster asynchronously, aka "fire and forget".
+    /// This function is non-blocking.
+    /// - Parameter message: The ``KafkaProducerMessage`` that is sent to the KafkaCluster.
+    /// - Returns: Unique message identifier matching the `id` property of the corresponding ``KafkaAcknowledgedMessage``
+    @discardableResult
+    public func sendAsync(message: KafkaProducerMessage) throws -> UInt {
+        try failIfProducerShutDown()
```
FranzBusch, line 169: Same here

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,246 @@
...
+        @convention(c) (OpaquePointer?, UnsafePointer<rd_kafka_message_t>?, UnsafeMutableRawPointer?) -> Void
+    ) = { _, messagePointer, producerPointer in
+        guard let producerPointer = producerPointer else {
+            fatalError("Could not resolve reference to KafkaProducer instance")
+        }
+        // Get a reference to the KafkaProducer without decrementing its reference count
+        let producer = Unmanaged<KafkaProducer>.fromOpaque(producerPointer).takeUnretainedValue()
+
+        guard let messagePointer = messagePointer else {
+            producer.logger.error("Could not resolved acknowledged message")
```
FranzBusch, line 216: Nit
```suggestion
            producer.logger.error("Could not resolve acknowledged message")
```

Sources/SwiftKafka/KafkaAcknowledgedMessage.swift
```diff
@@ -0,0 +1,70 @@
...
+    /// The offset of the message in its partition.
+    public let offset: Int64
+
+    /// Initialize `KafkaAckedMessage` from `rd_kafka_message_t` pointer.
+    init?(messagePointer: UnsafePointer<rd_kafka_message_t>, id: UInt? = nil) {
+        self.id = id
+
+        let rdKafkaMessage = messagePointer.pointee
+        guard rdKafkaMessage.err.rawValue == 0 else {
+            return nil
```
FranzBusch, line 41: Can we actually throw the Kafka error here?

Sources/SwiftKafka/KafkaAcknowledgedMessage.swift
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        guard let topic = String(validatingUTF8: rd_kafka_topic_name(rdKafkaMessage.rkt)) else {
+            return nil
```
FranzBusch, line 45: We should probably fatalError here

Sources/SwiftKafka/KafkaAcknowledgedMessage.swift
```diff
@@ -0,0 +1,70 @@
...
+                start: keyPointer,
+                count: rdKafkaMessage.key_len
+            )
+            self.key = .init(bytes: keyBufferPointer)
+        } else {
+            self.key = nil
+        }
+
+        guard let valuePointer = rdKafkaMessage.payload else {
+            return nil
```
FranzBusch, line 62: We should also fatalError here

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,246 @@
...
+            return
+        }
+
+        let messageID = messagePointer.pointee._private
+        guard let message = KafkaAcknowledgedMessage(messagePointer: messagePointer, id: UInt(bitPattern: messageID)) else {
```
FranzBusch, line 221: I already commented on the `KafkaAcknowledgedMessage` that we should make the `init` throwing and not return `nil`. Furthermore, we should also surface this error in the `AsyncSequence` otherwise users will never know what went wrong. What about having a `Result` type as the element of the `AsyncSequence` stream and we also introduce a `KafkaAcknowledgedMessageError` that contains the id of the message the actual error?

FranzBusch (commented):
I think these are the last round of comments that I have. Looks very good already!

Sources/SwiftKafka/KafkaProducer.swift
felixschlegel: Oh, must have overlooked it in the documentation 😅

FranzBusch (commented):
Looks good. Would love if @Lukasa can give this a look to check that we are handling the C inter-op correctly.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,268 @@
...
+    /// Method to shutdown the ``KafkaProducer``.
+    ///
+    /// This method flushes any buffered messages and waits until a callback is received for all of them.
+    /// Afterwards, it shuts down the connection to Kafka and cleans any remaining state up.
+    public func shutdownGracefully() throws {
+        switch self.state {
+        case .started:
+            self.state = .shuttingDown
+            self._shutDownGracefully()
+        case .shuttingDown, .shutDown:
```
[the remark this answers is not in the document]
felixschlegel: Done ✅

Sources/SwiftKafka/KafkaAcknowledgedMessageError.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Error caused by the Kafka cluster when trying to process a message produced by ``KafkaProducer``.
+public struct KafkaAcknowledgedMessageError: Error {
```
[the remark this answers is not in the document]
felixschlegel: https://github.com/swift-server/swift-kafka-gsoc/issues/25

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,279 @@
...
+            return
+        }
+    }
+
+    private func _shutDownGracefully() {
+        // Wait 10 seconds for outstanding messages to be sent and callbacks to be called
+        rd_kafka_flush(self.kafkaHandle, 10000)
```
Lukasa, line 170: We really shouldn't do this on the user thread: users are going to be _very_ surprised that their thread hangs for 10s.

Lukasa, line 170: Do we know that no further work can come in after this point?

Sources/SwiftKafka/KafkaAcknowledgedMessage.swift
```diff
@@ -0,0 +1,78 @@
...
+    /// The topic that the message was sent to.
+    public let topic: String
+    /// The partition that the message was sent to.
+    public let partition: Int32
+    /// The key of the message.
+    public let key: ByteBuffer?
+    /// The body of the message.
+    public let value: ByteBuffer
+    /// The offset of the message in its partition.
+    public let offset: Int64
```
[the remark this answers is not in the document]
felixschlegel: Not really. I initially thought that our users should not change anything about the data we provide them. But indeed, there is no reason why we should stop them from doing that.

Sources/SwiftKafka/KafkaError.swift
```diff
@@ -12,7 +12,25 @@
 //
 //===----------------------------------------------------------------------===//
 
-public struct KafkaError: Error {
+import Crdkafka
+
+struct KafkaError: Error {
```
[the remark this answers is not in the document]
felixschlegel: @FranzBusch and I are still unsure how we want to expose `Error`s in the public API. We, therefore, concluded that we do not want to expose this type until we have found a solution.

Sources/SwiftKafka/KafkaAcknowledgedMessage.swift
```diff
@@ -0,0 +1,78 @@
...
+        guard let valuePointer = rdKafkaMessage.payload else {
+            fatalError("Could not resolve payload of acknowledged message")
+        }
+
+        let valueBufferPointer = UnsafeRawBufferPointer(start: valuePointer, count: rdKafkaMessage.len)
+        self.value = .init(bytes: valueBufferPointer)
+
+        self.offset = Int64(rdKafkaMessage.offset)
+
+        guard rdKafkaMessage.err.rawValue == 0 else {
```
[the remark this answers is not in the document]
felixschlegel: Good point!

Sources/SwiftKafka/KafkaAcknowledgedMessage.swift
```diff
@@ -0,0 +1,78 @@
...
+            let keyBufferPointer = UnsafeRawBufferPointer(
+                start: keyPointer,
+                count: rdKafkaMessage.key_len
+            )
+            self.key = .init(bytes: keyBufferPointer)
+        } else {
+            self.key = nil
+        }
+
+        guard let valuePointer = rdKafkaMessage.payload else {
```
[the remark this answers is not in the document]
felixschlegel: Thank you for pointing that out! I have double-checked and tested what happens when sending a message with `length=0`. This results in `rdKafkaMessage.len` being 0, but the `rdKafkaMessage.payload` pointer still not being a Null-pointer. So I guess throwing an error when receiving a Null-pointer seems reasonable — however, it is debatable if the error should be a regular `Error` or a `fatalError`. I'll leave that up to you and @FranzBusch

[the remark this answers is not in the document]
Lukasa: I'm inclined to say we don't need to throw an error at all: we can just wrap this in an `UnsafeBufferPointer` which will tolerate being zero-length.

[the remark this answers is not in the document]
felixschlegel: Even better!

Sources/SwiftKafka/KafkaProducer.swift
felixschlegel: The 10 seconds were rather chosen arbitrarily — we could potentially have something like an outgoing message queue and wait until all outstanding messages have been acknowledged. However, this could take forever.

felixschlegel:
Hey @Lukasa,

First of all, thanks a lot for your review! I have applied your changes and replied to your conversations.
Please let me know what you think and if there is something I misunderstood.

Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -75,9 +75,28 @@ public struct KafkaConfig: Hashable, Equatable {
...
+            rd_kafka_conf_set_dr_msg_cb(
+                self.pointer,
+                callback
+            )
+        }
+
+        func setOpaque(opaque: AnyObject) {
+            rd_kafka_conf_set_opaque(
+                self.pointer,
+                Unmanaged.passUnretained(opaque).toOpaque()
```
FranzBusch, line 90: I find this more confusing than before that the `setOpaque` method is now hiding the `passUnretained` from the caller. I think @Lukasa meant something slightly different here where you are going to have an object that handles the `passUnretained`/`takeUnretained` for you and you just have to keep a reference to that object.

However, @Lukasa can probably expand more on this.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,289 @@
...
+            // Wait 10 seconds for outstanding messages to be sent and callbacks to be called
+            self.client.withKafkaHandlePointer { handle in
+                rd_kafka_flush(handle, 10000)
+            }
+        }
+
+        for (_, topicHandle) in self.topicHandles {
+            rd_kafka_topic_destroy(topicHandle)
+        }
+        self.pollTask.cancel()
```
FranzBusch, line 184: This doesn't seem right to me. We are flushing on a `Task` and concurrently destroying all the topic handles. Is that correct?

FranzBusch (commented):
Looks really good already. I think there is only one thing left where I would like to get @Lukasa opinion again.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,289 @@
...
+            logger: self.logger
+        )
+
+        self.config.setDeliveryReportCallback(callback: self.deliveryReportCallback)
+        self.config.setOpaque(opaque: self.callbackOpaque)
+
+        self.client = try KafkaClient(type: .producer, config: self.config, logger: self.logger)
+
+        // Poll Kafka every millisecond
+        self.pollTask = Task { [client] in
```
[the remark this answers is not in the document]
felixschlegel: We could make use of `NIOAsyncSequenceProducer` and its `Delegate` to trigger polling.

(I don't know if you are happy with how I implemented it in https://github.com/swift-server/swift-kafka-gsoc/pull/24, but if we find a good way to do it we can implement it here as well)

Sources/SwiftKafka/KafkaProducer.swift
felixschlegel: Uhhh yeah, *dumb* mistake — what do you think of getting rid of the `Task` and wrapping this blocking call with `await`? I guess that is what @Lukasa meant. Furthermore, we should think about if we even need a `shutDownGracefully()` method when we do not keep a reference to `self` in our `KafkaProducer`. We could then just make use of `deinit`. What do you think?

Sources/SwiftKafka/KafkaProducer.swift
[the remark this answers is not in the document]
FranzBusch: I don't think using an `AsyncSequence` for this is good. I just think we need a user `Task` to call a method on this actor like `run` and that method suspends until shutdown happens.

Sources/SwiftKafka/KafkaConfig.swift
felixschlegel: Ahh ok, I think I understand it now, please check in my latest commit!

Sources/SwiftKafka/KafkaProducer.swift
felixschlegel: We could potentially leave out the `flush` function call. However, this could lead to some queued messages getting lost without the user noticing. As far as I understand, the Kafka C library manages a message queue for outgoing messages. We can only put messages into this queue without influencing when they are sent. Therefore this is a safety mechanism to avoid messages in the queue getting lost when we close the producer.

Sources/SwiftKafka/KafkaClient.swift
```diff
@@ -45,12 +45,15 @@ final class KafkaClient {
...
 
-        guard let handle = rd_kafka_new(
-            clientType,
-            config.pointer,
-            errorChars,
-            KafkaClient.stringSize
-        ) else {
+        guard let handle = self.config.withDuplicatePointer({ [errorChars, clientType] duplicateConfig in
+            // Duplicate because rd_kafka_new takes ownership of the pointer and frees it.
+            rd_kafka_new(
```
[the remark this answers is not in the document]
felixschlegel: Thanks for pointing that out!

Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -117,4 +132,36 @@ public struct KafkaConfig: Hashable, Equatable {
 
         try self._internal.set(value, forKey: key)
     }
+
+    /// Define a function that is called upon every message acknowledgement.
+    /// - Parameter callback: C compatible function.
+    mutating func setDeliveryReportCallback(
+        callback: @escaping (@convention(c) (OpaquePointer?, UnsafePointer<rd_kafka_message_t>?, UnsafeMutableRawPointer?) -> Void)
```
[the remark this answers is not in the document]
felixschlegel: Yes! Do you think we should go with `AnyObject` here and do the `Unmanaged` call inside of this `func`? By doing this we could leave out unmanaged calls in our `KafkaProducer` entirely!

felixschlegel:
Hey @Lukasa ,

thanks again for re-reviewing my code! I hope that this new implementation is what you have thought of in the first place!
I have also answered all your comments in this PR. Thanks a lot!

Sources/SwiftKafka/KafkaProducer.swift
FranzBusch: It might make sense to let the user define that deadline. Instead of us fixing it to 10s

felixschlegel: Yes, let's do this! We can still stick with 10s as a default value

Sources/SwiftKafka/KafkaProducer.swift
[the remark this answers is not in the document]
felixschlegel: https://github.com/swift-server/swift-kafka-gsoc/issues/26

Sources/SwiftKafka/KafkaProducer.swift
felixschlegel: Have force-pushed a `timeout` parameter for `shutDownGracefully`, just to let you know

Sources/SwiftKafka/KafkaConfig.swift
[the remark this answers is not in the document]
felixschlegel: I thought about it a bit more and I think we can run into issues when trying to replace the`UnsafeMutableRawPointer` with `AnyObject` here in order to move the `Unmanaged` unwrapping to `KafkaConfig`. We have to keep in mind that C closures cannot capture context, so something like...

```swift
mutating func setDeliveryReportCallback(callback: @escaping (@convention(c) (OpaquePointer?, UnsafePointer<rd_kafka_message_t>?, AnyObject?) -> Void) {
  // Copy-on-write mechanism
  if !isKnownUniquelyReferenced(&(self._internal)) {
    self._internal = self._internal.createDuplicate()
  }
  
  self._internal.setDeliveryReportCallback { kafkaHandle, messagePointer, opaquePointer
    let opaqueObject = Unmanaged<AnyObject>.fromOpaque(opaquePointer).takeUnretainedValue()
    callback(kafkaHandle, messagePointer, opaqueObject) // 🛑 callback cannot be captured in C closure
  }
}
```

...would not work — or am I missing something here?

felixschlegel:
Hey @Lukasa, Hey @FranzBusch,

I have implemented additional changes that try to obscure the entire `Unmanaged` logic inside the `KafkaConfig`.
Please take a look at the individual commits:

 [9ce49e2](https://github.com/swift-server/swift-kafka-gsoc/pull/21/commits/9ce49e2c3892b4dd922da29e40f2d34f8bd304d3) is an implementation I worked out together with @FranzBusch and your ideas, @Lukasa 
 
 [834f243](https://github.com/swift-server/swift-kafka-gsoc/pull/21/commits/834f24396156da94883eb1fb5b971c1bce3cbfce) is an experimental simplification of what we worked out yesterday
 
 Please let me know what you think!

Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -22,19 +22,19 @@ import Crdkafka
 /// [all available configurations](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md).
 public struct KafkaConfig: Hashable, Equatable {
     private final class _Internal: Hashable, Equatable {
-        /// Pointer to the `rd_kafka_conf_t` object managed by `librdkafka`
+        /// Pointer to the `rd_kafka_conf_t` object managed by `librdkafka`.
         private(set) var pointer: OpaquePointer
```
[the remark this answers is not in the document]
Lukasa: Do you mean only within the `struct` or only within the `class`?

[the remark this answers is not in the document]
felixschlegel: The `pointer` property is only accessed within the `_Internal` `class`. I double-checked and realised that we can even replace `private(set)` with `private` here.

Sources/SwiftKafka/KafkaConfig.swift
```diff
@@ -75,9 +94,48 @@ public struct KafkaConfig: Hashable, Equatable {
...
+                let opaque = Unmanaged<CapturedClosure>.fromOpaque(opaquePointer).takeUnretainedValue()
+
+                let actualCallback = opaque.closure
+                actualCallback(messagePointer)
+            }
+
+            rd_kafka_conf_set_dr_msg_cb(
+                self.pointer,
+                callbackWrapper
+            )
```
FranzBusch, line 127: I think there might be a potential race if you are setting the callback a second time or nilling it out. The race is that you are setting `self.opaque` first which is going to release the current `CapturedClosure` and only at the very end you are setting the new callback in librdkafka. I think it is better if we would do the other way around and only override `self.opaque` after we set the new callback.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,21 +14,264 @@
...
+    /// The ``KafkaTopicConfig`` used for newly created topics.
     private let topicConfig: KafkaTopicConfig
+    /// A logger.
+    private let logger: Logger
+    /// Dictionary containing all topic names with their respective `rd_kafka_topic_t` pointer.
+    private var topicHandles: [String: OpaquePointer]
+
+    // We use implicitly unwrapped optionals here as these properties need to access self upon initialization
+    /// The configuration object of the producer client.
+    private var config: KafkaConfig!
```
FranzBusch, line 82: I think we can get rid of this. The config is already retained in the `KafkaClient` and we never use it outside the init here

FranzBusch (commented):
Two last small comments. Otherwise this looks great!

FranzBusch (approved):
LGTM! Thanks for incorporating all the changes along the way

PR #24, opened 2022-09-06
Implementation of the `KafkaConsumer`

felixschlegel:
## Modifications:

* implemented the `KafkaConsumer`
* `KafkaConsumer` has two public initializers:

  ```swift
  // Subscribe to topics as part of a consumer group
  KafkaConsumer(
      topics: [String],
      groupID: String,
      config: KafkaConfig = KafkaConfig(),
      logger: Logger
  ) async throws {}
  
  // Assign consumer to a particular topic partition
  KafkaConsumer(
      topic: String,
      partition: KafkaPartition,
      offset: Int64 = Int64(RD_KAFKA_OFFSET_END),
      config: KafkaConfig = KafkaConfig(),
      logger: Logger
  ) async throws {
  ```
* incoming messages can be read using `KafkaConsumer`'s `messages` `AsyncSequence` that has **backpressure** implemented
* added the ability for manual offset management by implementing the `commitSync()` method for `KafkaConsumer`
* implemented **Integration Tests**

## Result:

Users can consume messages from a Kafka cluster.


Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,418 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+import struct Foundation.UUID // TODO: can we avoid this Foundation import??
+import Logging
+import NIOCore
+
+/// `NIOAsyncSequenceProducerDelegate` implementation handling backpressure for ``KafkaConsumer``.
+actor ConsumerMessagesAsyncSequenceDelegate: NIOAsyncSequenceProducerDelegate {
```
felixschlegel, line 21: Went with an `actor` as we need to be thread-safe and have this `weak var consumer`. However, we could think about a different approach here.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,418 @@
...
+        }
+    }
+
+    public func makeAsyncIterator() -> ConsumerMessagesAsyncIterator {
+        return ConsumerMessagesAsyncIterator(wrappedIterator: self.wrappedSequence.makeAsyncIterator())
+    }
+}
+
+/// Receive messages from the Kafka cluster.
+public actor KafkaConsumer {
```
felixschlegel, line 65: Again, chose an `actor` here to have out-of-the-box thread safety for our `NIOAsyncSequenceProducer`. Please double-check how I implemented the connection between the `NIOAsyncSequenceProducer` and the `KafkaConsumer`, maybe there is a better approach.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,418 @@
...
+
+    /// States that the ``KafkaConsumer`` can have.
+    private enum State {
+        /// The ``KafkaConsumer`` has started and is ready to use.
+        case started
+        /// The ``KafkaConsumer`` has been closed and does not receive any more messages.
+        case closed
+    }
+
+    // TODO: do we want to allow users to subscribe / assign to more topics during runtime?
```
felixschlegel, line 77: Do we want to allow users to subscribe/assign to more topics during runtime?

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,418 @@
...
+    /// - Parameter logger: A logger.
+    public convenience init(
+        topics: [String],
+        groupID: String,
+        config: KafkaConfig = KafkaConfig(),
+        logger: Logger
+    ) async throws {
+        // TODO: make prettier
+        // TODO: use inout parameter here?
+        var config = config
```
felixschlegel, line 157: I am not really happy with how I am modifying the `config` here. What do you think?

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,418 @@
...
+        case .closed:
+            return
+        }
+    }
+
+    /// Subscribe to the given list of `topics`.
+    /// The partition assignment happens automatically using `KafkaConsumer`'s consumer group.
+    /// - Parameter topics: An array of topic names to subscribe to.
+    private func subscribe(topics: [String]) throws {
+        // TODO: is this state needed for a method that is only invoked upon init?
```
felixschlegel, line 214: Currently, `subscribe` is only invoked by `init`. Thus we could potentially get rid of the state. But this totally depends on if we want to allow users to `subscribe/assign` to more topics during runtime.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,418 @@
...
+    /// Assign the``KafkaConsumer`` to a specific `partition` of a `topic`.
+    /// - Parameter topic: Name of the topic that this ``KafkaConsumer`` will read from.
+    /// - Parameter partition: Partition that this ``KafkaConsumer`` will read from.
+    /// - Parameter offset: The topic offset where reading begins. Defaults to the offset of the last read message.
+    private func assign(
+        topic: String,
+        partition: KafkaPartition,
+        offset: Int64 = Int64(RD_KAFKA_OFFSET_END)
+    ) throws {
+        // TODO: is this state needed for a method that is only invoked upon init?
```
felixschlegel, line 248: Currently, `assign` is only invoked by `init`. Thus we could potentially get rid of the state. But this totally depends on if we want to allow users to `subscribe/assign` to more topics during runtime.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    /// Request a new message from the Kafka cluster.
+    /// This method blocks for a maximum of `timeout` milliseconds.
+    /// - Parameter timeout: Maximum amount of milliseconds this method waits for a new message.
+    /// - Returns: A ``KafkaConsumerMessage`` or `nil` if there are no new messages.
+    func poll(timeout: Int32 = 100) async throws -> KafkaConsumerMessage? {
+        // TODO: clock API
+        // TODO: ideally: make private
+        // TODO: is this state needed here?
```
felixschlegel, line 303: 1. This is a good candidate for `Clock API` — shall I create a ticket?
2. Again, this function is not meant to be used from outside the `KafkaConsumer` — we could therefore think about omitting `state` here as well

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,418 @@
...
+        switch self.state {
+        case .started:
+            break
+        case .closed:
+            throw KafkaError(description: "Trying to invoke method on consumer that has been closed.")
+        }
+        try await self._commitSync(message)
+    }
+
+    // TODO: commit multiple messages at once -> different topics + test
```
felixschlegel, line 354: What do you think about refactoring this function into a function that takes an array of `KafkaConsumerMessage` as a parameter and commits all of them at once?

This would then always commit the highest offset for each topic.

Sources/SwiftKafka/KafkaConsumerMessage.swift
```diff
@@ -0,0 +1,70 @@
...
+        self.value = .init(bytes: valueBufferPointer)
+
+        self.offset = Int64(rdKafkaMessage.offset)
+
+        guard rdKafkaMessage.err == RD_KAFKA_RESP_ERR_NO_ERROR else {
+//            var errorStringBuffer = ByteBuffer(bytes: valueBufferPointer)
+//            let errorString = errorStringBuffer.readString(length: errorStringBuffer.readableBytes)
+
+            // TODO: what to do with error string?
+            // TODO: handle errors here or in consumer?
```
felixschlegel, line 66: Handle errors here or in `KafkaConsumer`?

Tests/IntegrationTests/SwiftKafkaTests.swift
```diff
@@ -0,0 +1,222 @@
...
+
+    // TODO: also test concurrently?
+
+    // MARK: - Helpers
+
+    private static func resetConsumerOffset(_ consumer: KafkaConsumer) async throws {
+        let start = Date()
+        let timeout = 10.0
+        while Date() < start + timeout {
+            _ = try await consumer.poll(timeout: 1000) // TODO: poll needs to be private, do something else here
```
felixschlegel, line 184: Ideally, we would be able to make `poll()` `private` and have an `AsyncSequence` with a **timeout** here. Though afaik, there is no such thing so we would have to implement it ourselves.

What do you think of this approach?

https://forums.swift.org/t/running-an-async-task-with-a-timeout/49733

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,418 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+import struct Foundation.UUID // TODO: can we avoid this Foundation import??
```
felixschlegel, line 16: Although `assign` is not using consumer groups, we have to specify a `group.id` for every `KafkaConsumer`. That's why I use `UUID` for generating unique group ids for assignments. However, I am not really happy with importing `Foundation` just to have `UUID`. Do you have any idea how we can circumvent this?

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,418 @@
...
+import Logging
+import NIOCore
+
+/// `NIOAsyncSequenceProducerDelegate` implementation handling backpressure for ``KafkaConsumer``.
+actor ConsumerMessagesAsyncSequenceDelegate: NIOAsyncSequenceProducerDelegate {
+    weak var consumer: KafkaConsumer?
+
+    // TODO: is it ok to block here -> we can make poll non-async and get rid of the task here
+    nonisolated func produceMore() {
```
felixschlegel, line 25: Is it ok if `producerMore()` is blocking? We could then get rid of the `Task` etc.

mr-swifter:
Hello @felixschlegel ! 

I was looking at the PR have one small question, just curious. 

Do you have plans to address that now some ```lifrdkafka```'s blocking functions are called from Tasks? I see, in particular:
- ```rd_kafka_commit``` from ```commitSync```/```_commitSync```
- ```rd_kafka_consumer_poll``` from ```produceMore``` 

Or it's beyond the scope? That would be very nice to make the client truly behave async. 

I would be happy to discuss how that can be addressed.

felixschlegel:
> Hello @felixschlegel !
> 
> I was looking at the PR have one small question, just curious.
> 
> Do you have plans to address that now some `lifrdkafka`'s blocking functions are called from Tasks? I see, in particular:
> 
> * `rd_kafka_commit` from `commitSync`/`_commitSync`
> * `rd_kafka_consumer_poll` from `produceMore`
> 
> Or it's beyond the scope? That would be very nice to make the client truly behave async.
> 
> I would be happy to discuss how that can be addressed.

Hey @mr-swifter,

Nice to hear that you are getting involved in developing this package!

Thanks for pointing out that the `Tasks` might get us into trouble. In [this comment](https://github.com/swift-server/swift-kafka-gsoc/pull/24#discussion_r967865567), I already addressed the desire to get rid of the `Task` in the `produceMore` method.

It is possible to make the `commitSync` method behave async in a non-blocking manner. However, we would only be able to know if the commit was successfully scheduled and not if the commit itself was successful.

I hope I understood your comment correctly. Please let me know what you think!

mr-swifter:
Hi @felixschlegel ! 

Yep, it seems I missed this comment when was looking at changes. It's good if you have some plans to fix this! 

Regarding ```commitSync``` I mean that it would be nice to make the call behave that ```await consumer.commitSync``` will be returned back only when commit is send and has been replied by kafka, so client can return proper result to user, but meantime the Task where ```commitSync``` happens will be properly suspended, but not sitting on blocking call, so the thread can execute other tasks meanwhile.

I do something like this at my version, maybe not perfect, but works in my case and provide desired behaviour:

```
func commitSync(...) async { 
    await withCheckedContinuation { continuation in
        kafkaDispatchQueue.async { // shedule executution to separate queue managed by my kafka client package outside of cooperative pool 
            let result = rd_kafka_commit(...) // or any other blocking or timeout-based blocking call like rd_kafka_consumer_poll
            /// wrap-up the result if needed 
            continuation.resume(returning: result)
        }
    }
}

...
let result = await consumer.commitSync()
/// here we know the result of commit
```

So, basically I use separate execution queue for scheduling blocking calls and use continuation to resume task execution when ```librdkafka``` blocking function is over. 
At the moment it's single queue, but I see maybe I will need to migrate to a few queue for different purposes (like poll vs commits vs transactions etc).

FranzBusch:
@mr-swifter Thanks for raising this. I fully agree with you that we should not block any of the cooperative thread pools threads here since it can lead to various problems and the proper thing to do is actually back the consumer by a queue and dispatch every call onto this queue. So that we only block the thread backing the queue and just take a continuation that will be resumed once the blocking call finished.

The interesting challenge here is going to be handling cancellation; however, in the beginning we could just punt that and implement that at a later point. Having a queue is already a bigger refactor and brings us a lot of benefits.

felixschlegel:
Hey @mr-swifter , Hey @FranzBusch ,

thank you for your comments! I have implemented a single serial `DispatchQueue` that takes care of all our blocking methods. Let me know what you think!

FranzBusch (commented):
Looks really good already, I left a few comments inline. There are a few larger points which I would like us to track in issues.
1. We should think again if we may want to split this type into `KafkaSinglePartitionConsumer` and `KafkaGroupConsumer`
2. librdkafka exposes a callback based API for consuming messages which ought to be faster than the poll based one. We should investigate this

mr-swifter:
Great stuff! Thanks @felixschlegel , @FranzBusch !

Sources/SwiftKafka/KafkaConsumerMessage.swift
felixschlegel: We could move it up but we would have to move up the `valuePointer` and the `valueBufferPointer` as well — I'll implement it in my next commit

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,434 @@
...
+    /// Serial queue used to run all blocking operations. Additionally ensures that no data races occur.
+    private let serialQueue: DispatchQueue
+
+    /// Type of the values returned by the ``messages`` sequence.
+    private typealias Element = Result<KafkaConsumerMessage, Error> // TODO: replace with a more specific Error type
+    private let messagesSource: NIOAsyncSequenceProducer<
+        Element,
+        ConsumerMessagesAsyncSequence.HighLowWatermark,
+        ConsumerMessagesAsyncSequenceDelegate
+    >.Source
```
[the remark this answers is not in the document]
felixschlegel: Can we postpone this topic?

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,434 @@
...
+        }
+    }
+
+    /// Helper function that enqueues a task with a checked throwing continuation into the ``KafkaConsumer``'s serial queue.
+    private func serializeWithThrowingContinuation<T>(_ body: @escaping (CheckedContinuation<T, Error>) -> Void) async throws -> T {
+        try await withCheckedThrowingContinuation { continuation in
+            self.serialQueue.async {
+                body(continuation)
+            }
+        }
```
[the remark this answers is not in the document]
felixschlegel: https://github.com/swift-server/swift-kafka-gsoc/issues/33

felixschlegel:
Hey @FranzBusch,

I have implemented your suggested changes and created new issues for some of your comments. Please let me know what you think!

With the new topic creation method, I get this error when testing `commitSync` in the `IntegrationTests`:

```bash
%5|1663939258.150|PARTCNT|rdkafka#consumer-9| [thrd:main]: Topic 6B1E56D4-3387-4CEA-AC41-FB92471FBAFD partition count changed from 1 to 0
%4|1663939258.189|COMMITFAIL|rdkafka#consumer-9| [thrd:main]: Offset commit (unassigned partitions) failed for 1/1 partition(s) in join-state init: Broker: Unknown topic or partition: 6B1E56D4-3387-4CEA-AC41-FB92471FBAFD[0]@10(Broker: Unknown topic or partition)
```

However, the test still succeeds. Just wanted to let you know, maybe you have an idea what's going on here!

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,393 @@
...
+    func produceMore() {
+        self.serialQueue.async {
+            guard !self.closed else {
+                return
+            }
+
+            let messageResult: Element
+            do {
+                guard let message = try self.poll() else {
+                    self.produceMore()
```
felixschlegel, line 256: @FranzBusch 
Is this behaviour correct or will the `NIOAsyncSequence` invoke `produceMore()` again anyway?

FranzBusch: I think for the shape that we are doing here it is okay to call `self.produceMore()` again. The `NIOAsyncSequenceProducer` will only call `produceMore` once when an element has been consumed and you need to make sure that at some point that leads to a new message.

FranzBusch (approved):
Let's get this merged. I think we made some good progress here on an initial consumer and opened a few issues to improve the API.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -0,0 +1,393 @@
...
+        let backpressureStrategy = ConsumerMessagesAsyncSequence.HighLowWatermark(
+            lowWatermark: 5,
+            highWatermark: 10
+        )
+
+        let messagesSequenceDelegate = ConsumerMessagesAsyncSequenceDelegate { [weak self] in
+            self?.produceMore()
+        } didTerminateClosure: { [weak self] in
+            self?.close()
+        }
```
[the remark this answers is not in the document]
felixschlegel: https://github.com/swift-server/swift-kafka-gsoc/issues/37

Issue #25, opened 2022-09-07
Improve public interface for `KafkaAcknowledgedMessageError`


felixschlegel:
Currently `KafkaAcknowledgedMessageError` uses a preliminary public interface. This shall be changed to a more **user-friendly** and **descriptive** version.

> Franz: we do not want to expose `.rawValue` as `Int32`

Issue #29, opened 2022-09-14
Improve general public interface of Errors


felixschlegel:
**Discovery Task:** What `public` error types do we want to expose in this package?

Having figured that out, please also **implement** those error types.

> **Note**
> Please also include those new error types in the DocC documentation of their throwing function using the `Throws` aside:
> `/// - Throws:`

PR #36, opened 2022-09-25
Improved implementation of `KafkaError` and `KafkaAcknowledgedMessageError`

felixschlegel:
Closes issues:

#25, #29 

## Motivation:

The general idea is to have a **single error type (namely `KafkaError`)** that is used for all errors in the program. 
However, I still kept the `KafkaAcknowledgedMessageError` as we somehow need to provide our users with a convenient way to access the `messageID` of a failed message acknowledgement.

The challenge I faced in this PR is that the C library comes with its error types that shall not be exposed publicly. My solution is to wrap the C library's errors inside a dedicated `case` of the `KafkaError` type.

### DIFF
--- a/Sources/SwiftKafka/KafkaAcknowledgedMessage.swift
+++ b/Sources/SwiftKafka/KafkaAcknowledgedMessage.swift
@@ -33,6 +33,7 @@ public struct KafkaAcknowledgedMessage: Hashable {
     public var offset: Int64
 
     /// Initialize ``KafkaAcknowledgedMessage`` from `rd_kafka_message_t` pointer.
+    /// - Throws: A ``KafkaAcknowledgedMessageError`` for failed acknowledgements or malformed messages.
     init(messagePointer: UnsafePointer<rd_kafka_message_t>, id: UInt) throws {
         self.id = id
 
--- a/Sources/SwiftKafka/KafkaAcknowledgedMessage.swift
+++ b/Sources/SwiftKafka/KafkaAcknowledgedMessage.swift
@@ -41,15 +41,15 @@ public struct KafkaAcknowledgedMessage: Hashable {
         let valueBufferPointer = UnsafeRawBufferPointer(start: rdKafkaMessage.payload, count: rdKafkaMessage.len)
         self.value = ByteBuffer(bytes: valueBufferPointer)
 
-        guard rdKafkaMessage.err.rawValue == 0 else {
+        guard rdKafkaMessage.err == RD_KAFKA_RESP_ERR_NO_ERROR else {
             var errorStringBuffer = self.value
             let errorString = errorStringBuffer.readString(length: errorStringBuffer.readableBytes)
 
-            throw KafkaAcknowledgedMessageError(
-                rawValue: rdKafkaMessage.err.rawValue,
-                description: errorString,
-                messageID: self.id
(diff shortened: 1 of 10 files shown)


FranzBusch (commented):
I think this is off to a great start; however, I would love if we can change the shape of the Errors a bit. We have developed quite a unique pattern in the NIO world over the last projects that allow us to have a forward evolvable type and capture as much information as possible.

You can basically copy that as is and just slightly adapt the naming and add all the error codes and func for the errors we have. 
```swift
public struct FooError: Error, Hashable, CustomStringConvertible {
    private var backing: Backing

    /// Represents the kind of error that was encountered.
    public var code: ErrorCode {
        get {
            self.backing.code
        }
        set {
            self.makeUnique()
            self.backing.code = newValue
        }
    }

    private var reason: String {
        self.backing.reason
    }

    private var file: String {
        self.backing.file
    }

    private var line: UInt {
        self.backing.line
    }

    public var description: String {
        "FooError.\(self.code): \(self.reason) \(self.file):\(self.line)"
    }

    private mutating func makeUnique() {
        if !isKnownUniquelyReferenced(&self.backing) {
            self.backing = self.backing.copy()
        }
    }

    public static func bar(
        reason: String, file: String = #fileID, line: UInt = #line
    ) -> FooError {
        return FooError(
            backing: .init(
                code: .bar reason: reason, file: file, line: line
            )
        )
    }
}

extension FooError {
    /// Represents the kind of an error.
    ///
    /// The same kind of error may be thrown from more than one place, for more than one reason. This type represents
    /// only a fairly high level kind of error: use the string representation of ``FooError`` to get more details
    /// about the specific cause.
    public struct ErrorCode: Hashable, Sendable, CustomStringConvertible {
        fileprivate enum BackingCode {
            case bar
        }

        fileprivate var backingCode: BackingCode

        fileprivate init(_ backingCode: BackingCode) {
            self.backingCode = backingCode
        }

        /// Docs
        public static let bar = ErrorCode(.bar)

        public var description: String {
            return String(describing: self.backingCode)
        }
    }
}

extension FooError {
    final class Backing: Hashable {
        var code: FooError.ErrorCode

        let reason: String

        let file: String

        let line: UInt

        fileprivate init(
            code: FooError.ErrorCode,
            reason: String,
            file: String,
            line: UInt
        ) {
            self.code = code
            self.reason = reason
            self.file = file
            self.line = line
        }

        // Only the error code matters for equality.
        static func ==(lhs: Backing, rhs: Backing) -> Bool {
            return lhs.code == rhs.code
        }

        func hash(into hasher: inout Hasher) {
            hasher.combine(self.code)
        }

        fileprivate func copy() -> Backing {
            return Backing(code: self.code, reason: self.reason, file: self.file, line: self.line)
        }
    }
}
```

Furthermore, I am also thinking about making the `KafkaAcknowledgedMessageError` just a wrapper around the `KafkaError` where it basically has two fields `messageID` and `error`.

felixschlegel:
Hey @FranzBusch ,

Thank you for your review of this PR! I have implemented your suggestion accordingly. Please have a look 😄

Regarding our `static` error types: can we omit the `reason:` and `wrapper:` labels?

FranzBusch (approved):
I think this looks good for now! Let's merge it

Issue #9, opened 2022-06-24
Create `KafkaConfig` struct for global configurations


felixschlegel:
Build a struct called `KafkaConfig` that exposes a string-based dictionary that enables the user to configure their Kafka client.

Issue #10, opened 2022-06-24
Create `KafkaTopicConfig` struct for topic configurations


felixschlegel:
Build a struct called `KafkaTopicConfig` that exposes a string-based dictionary that enables the user to configure the new topic that is created by the `KafkaProducer` when a message is sent to a non-existing topic.

PR #42, opened 2022-11-20
Strongly-typed Configurations

felixschlegel:
Closes issues: #9, #10

## Motivation

This PR introduces the new `ConsumerConfig` and `ProducerConfig` that enable users to configure their `KafkaConsumer`/`KafkaProducer` using a strongly-typed `struct` rather than a string-based dictionary.

## Approach

* There are two `structs` called `ConsumerConfig` and `ProducerConfig`
* `ConsumerConfig` and `ProducerConfig` share a lot of the same configuration properties, thus I introduced a `protocol ClientConfig` that uses protocol extension to mimic an **abstract class** that takes care of all common config properties
* both configs contain a **dictionary** keeping track of all configuration string values that are required by `librdkafka`

> **Note:** These new configuration types are currently translated to the legacy configuration type.
> Replacing the legacy configuration type with the new configuration types will happen in **another PR.**

## TODO

- [x] DocC Documentation
- [x] New topic configuration `struct`
- [ ] Tests for new configs (are values being set correctly)
- [ ] Double-check if there are missing/redundant options (maybe only use official Kafka config options?)

## Open Questions

* Do we want to support `OAuth` with our strongly typed configs?
* Do we want to support setting SSL PEM keys in our strongly-typed configs?


Sources/SwiftKafka/New/ClientConfig.swift
```diff
@@ -0,0 +1,278 @@
...
+
+// TODO: dict add manually by key (unsafe api)?
+// TODO: support PEMs?
+// TODO: support OAuth?
+// TODO: DocC: inherit documentation?
+// TODO: DocC: take from lirbdkafka official documentation
+// TODO: Topic config -> see KafkaConfig in SwiftKafka
+// TODO: test that values get set accordingly
+// TODO: remove old config tests
+public protocol ClientConfig: Hashable, Equatable {
```
FranzBusch, line 23: I don't think this should be a protocol. We should rather use a struct here and hide the whole dictionary thing internally. Using a protocol for this would require us to either make the consumers generic or use an existential and both are not really great.

FranzBusch: Furthermore, what I would recommend is that we create a separate `ConsumerConfig` and `ProducerConfig` like you did and then have an internal method on them that can convert them to librdkafka config.

FranzBusch (commented):
Back from vacation and had a chance to look at the PR finally. In general, this goes into the right direction but I think we should model the API slightly differently.

felixschlegel:
Hey @FranzBusch,

I have implemented some changes since we last talked and would like to hear your feedback 😄 :

* a new `struct TopicConfiguration`, similar to `ConsumerConfig` and `ProducerConfig`
* DocC documentation (mostly copy-pasted from the official [librdkafka config documentation](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md))
* removed the `ClientConfig` and duplicated the common config options. However, introduced a new `protocol StringDictionaryRepresentable` for types that have a dictionary representation (not used as a type, but rather for uniformity)
* created a follow-up task for refactoring the package into using this new configuration API: #43 

There are still some things I am unsure about:

* I would like to test that everything gets set correctly. Should I write a test testing that each value is set correctly, or is that too redundant?
* Currently, we expose more or less all `librdkafka` configurations — do we want to narrow it down to the most basic configurations and expand this on user request / provide an unsafe API that allows pro-users to set arbitrary key-value options?
* librdkafka exposes configuration options regarding `OAuth` and SSL keys etc. — I have not tested this — does our package even intend to expose that as well?

I hope that I have explained my questions in enough detail. Please contact me if anything is unclear!

Season's greetings 🎅🏼

Felix

Sources/SwiftKafka/New/ConsumerConfig.swift
```diff
@@ -0,0 +1,456 @@
...
+
+    /// Client group id string. All clients sharing the same group.id belong to the same group.
+    public var groupID: String {
+        get { self.dictionary["group.id"] ?? "" }
+        set { self.dictionary["group.id"] = String(newValue) }
+    }
+
+    /// Client group session and failure detection timeout. The consumer sends periodic heartbeats (heartbeat.interval.ms) to indicate its liveness to the broker. If no hearts are received by the broker for a group member within the session timeout, the broker will remove the consumer from the group and trigger a rebalance. The allowed range is configured with the broker configuration properties group.min.session.timeout.ms and group.max.session.timeout.ms. Also see max.poll.interval.ms.
+    public var sessionTimeoutMs: UInt {
+        get { self.dictionary.getUInt("session.timeout.ms") ?? 45000 }
```
FranzBusch, line 28: Are all the default values here returned the ones that librdkafka uses?

FranzBusch (approved):
Thanks for the work. I think this looks quite good. I have some open concerns around some of the public types but let's move forward and reevaluate the public types at another point in time.

FranzBusch:
> I would like to test that everything gets set correctly. Should I write a test testing that each value is set correctly, or is that too redundant?
Currently, we expose more or less all librdkafka configurations — do we want to narrow it down to the most basic configurations and expand this on user request / provide an unsafe API that allows pro-users to set arbitrary key-value options?
librdkafka exposes configuration options regarding OAuth and SSL keys etc. — I have not tested this — does our package even intend to expose that as well?

Let's put all these open questions into GH issues for now

Sources/SwiftKafka/New/ConsumerConfig.swift
felixschlegel: Yes! I used [librdkafka/CONFIGURATION.md](https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md) for reference.

felixschlegel:
> > I would like to test that everything gets set correctly. Should I write a test testing that each value is set correctly, or is that too redundant?
> > Currently, we expose more or less all librdkafka configurations — do we want to narrow it down to the most basic configurations and expand this on user request / provide an unsafe API that allows pro-users to set arbitrary key-value options?
> > librdkafka exposes configuration options regarding OAuth and SSL keys etc. — I have not tested this — does our package even intend to expose that as well?
> 
> Let's put all these open questions into GH issues for now

Done!

For reference: #45 , #46

kylebrowning:
@felixschlegel is this ready to go?

FranzBusch:
@kylebrowning yep it was just waiting on me to merge it

kylebrowning:
Cool thanks, Im going to start using this in a related APNS project :)

PR #48, opened 2023-04-04
New `KafkaConfig` Refactoring

felixschlegel:
### Motivation:

Before this change, we had a `KafkaConfig` class and a `KafkaTopicConfig` class that were responsible for creating, updating and deleting an underlying `rd_kafka_(topic_)conf_t` object, which meant that every change to the `Kafka(Topic)Config` had to be reflected in the underlying `rd_kafka_(topic_)conf_t` object.

In this new approach, we have more lightweight structs, namely `struct Kafka(Topic, Producer, Consumer)Config` that only store configuration values without any `rd_kafka` logic running in the background.

We now only create `rd_(topic_)conf_t` objects right before using them with another `rd_kafka` function and keep configuration types in the Swift world for as long as possible.

### Modifications:

* create a new folder, `RDKafka` that contains Swift functions wrapping the standard `rd_kafka_conf_*` functions
* update README to include the new configuration types
* temporarily disable the `socketConnectionSetupTimeoutMs` config option because it is not yet supported by `librdkafka` on Linux
* invoke `rd_kafka_conf_new` in `KafkaClient` (through `RD` wrapper) 
* invoke `rd_kafka_topic_conf_new` in `KafkaProducer` (through `RD` wrapper) 
* rename `TopicConfig` -> `KafkaTopicConfig`
* rename `ProducerConfig` -> `KafkaProducerConfig`
* rename `ConsumerConfig` -> `KafkaConsumerConfig`
* delete `Kafka(Topic)Config` and all related tests

### Result:

* everything should work as before
* syntax for using a config in a Kafka client has slightly changed
* number of tests has decreased


Sources/SwiftKafka/Configuration/KafkaConsumerConfig.swift
```diff
@@ -192,10 +192,11 @@ public struct ConsumerConfig: Hashable, Equatable, StringDictionaryRepresentable
     }
 
     /// Maximum time allowed for broker connection setup (TCP connection setup as well SSL and SASL handshake). If the connection to the broker is not fully functional after this the connection will be closed and retried.
-    public var socketConnectionSetupTimeoutMs: UInt {
-        get { self.dictionary.getUInt("socket.connection.setup.timeout.ms") ?? 30000 }
-        set { self.dictionary["socket.connection.setup.timeout.ms"] = String(newValue) }
-    }
+    // TODO: fix issue with librdkafka version being too old on Linux
```
felixschlegel, line 195: This issue should be resolved once #49 is merged

Sources/SwiftKafka/Configuration/KafkaConfigEnums.swift
```diff
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 /// Collection of `enum` types used in the configuration structs this library provides.
-public struct ConfigEnums {
+public struct KafkaConfigEnums {
```
FranzBusch, line 16: Can we make this `public enum KafkaSharedConfiguration`

Sources/SwiftKafka/Configuration/StringDictionaryRepresentable.swift
FranzBusch, line 1: Do we still need this protocol?

Sources/SwiftKafka/RDKafka/RDKafkaConfig.swift
```diff
@@ -0,0 +1,103 @@
...
+// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+
+/// A collection of helper functions wrapping common `rd_kafka_conf_*` functions in Swift.
+struct RDKafkaConfig {
```
FranzBusch, line 18: Can we make this a class that holds the `rd_kafka_conf_new` pointer in an optional. If the optional is `some` we are going to destroy it in `deinit`. Additionally we offer a method that creates a new `rd_kafka` and sets the optional to `nil`. This way we can ensure in a single class that nothing breaks this.

Sources/SwiftKafka/RDKafka/RDKafkaConfig.swift
```diff
@@ -0,0 +1,103 @@
...
+
+    /// A Swift wrapper for `rd_kafka_conf_set_dr_msg_cb`.
+    /// Defines a function that is called upon every message acknowledgement.
+    /// - Parameter configPointer: An `OpaquePointer` pointing to the `rd_kafka_conf_t` object in memory.
+    /// - Parameter callback: A closure that is invoked upon message acknowledgement.
+    /// - Returns: A ``CapturedClosure`` object that must me retained by the caller as long as acknowledgements are received.
+    static func setDeliveryReportCallback(
+        configPointer: OpaquePointer,
+        _ callback: @escaping ((UnsafePointer<rd_kafka_message_t>?) -> Void)
+    ) -> CapturedClosure {
```
FranzBusch, line 73: Instead of returning a captured closure here we should be able to express this more safe in Swift. Above I already mentioned that I think we should have a method on `RDKafkaConfig` that returns an `rd_kafka`. We should probably have that method return `KafkaClient` instead and add an init to `KafkaClient` that takes both a `rd_kafka` and `CapturedClosure` to make sure this is never dropped.

Sources/SwiftKafka/RDKafka/RDKafkaTopicConfig.swift
```diff
@@ -0,0 +1,54 @@
...
+// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+
+/// A collection of helper functions wrapping common `rd_kafka_topic_conf_*` functions in Swift.
+struct RDKafkaTopicConfig {
```
FranzBusch, line 18: Can we apply the same here with the class instead of struct approach.

FranzBusch (changes requested):
Looks good already! Left some comments inline

Sources/SwiftKafka/Configuration/StringDictionaryRepresentable.swift
felixschlegel: No we don't, great catch!

Sources/SwiftKafka/RDKafka/RDKafkaConfig.swift
felixschlegel: I think our Package should **not** keep a reference to the `OpaquePointer` of the `rd_kafka_conf_t`. This is because in `KafkaClient`, we invoke [`rd_kafka_new`](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#a63d5cd86ab1f77772b2be170e1c09c24) which takes the `rd_kafka_conf_t` pointer we just created as an argument and assumes ownership of it (meaning it frees it once the new Kafka connection has been established). `RDKafkaConfig.createFrom` should therefore just serve as a wrapper for `rdkafka` C functions and not contain any more pointer ownership logic as this is handled by `rdkafka` itself already.

Sources/SwiftKafka/RDKafka/RDKafkaTopicConfig.swift
felixschlegel: (See comment about `RDKafkaConfig`)

FranzBusch (approved):

PR #52, opened 2023-05-04
Exclude some files from `soundness.sh`

felixschlegel:
## Motivation:

CI failed as our `soundness.sh` script detected unacceptable language in our repo that cannot be avoided.

## Modifications:

* excluded `CODE_OF_CONDUCT.md` from soundness check as it deliberately gives examples of what unacceptable language looks like
* excluded all `*Config.swift` files from soundness check as Kafka configuration properties rely on terminology which is considered unacceptable

## Result:

The soundness CI check succeeds.

### DIFF
--- a/scripts/soundness.sh
+++ b/scripts/soundness.sh
@@ -30,9 +30,22 @@ unacceptable_terms=(
     -e slav[e]
     -e sanit[y]
 )
-if git grep --color=never -i "${unacceptable_terms[@]}" > /dev/null; then
+
+# We have to exclude the code of conduct because it gives examples of unacceptable language.
+# We have to exclude *Config.swift files as they need to map to Kafka terminology
+# which is considered unacceptable by us.
+exclude_files=(
+    CODE_OF_CONDUCT.md
+    *Config.swift
+)
+for word in "${exclude_files[@]}"; do
+    exclude_files+=(":(exclude)$word")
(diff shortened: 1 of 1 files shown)


felixschlegel:

Tests/IntegrationTests/SwiftKafkaTests.swift
```diff
@@ -201,7 +201,7 @@ final class SwiftKafkaTests: XCTestCase {
         var messageIDs = Set<UInt>()
 
         for message in messages {
-            messageIDs.insert(try await producer.sendAsync(message))
+            try messageIDs.insert(await producer.sendAsync(message))
```
gjcairo, line 204: We usually prefer to disable `hoistTry` (and `hoistAwait`) to avoid moving `try`s and `await`s to the beginning of the line. It's generally not useful to lose clarity of what is throwing/async, and sometimes actually generates compiler errors because you can end up with malformed syntax.

You can disable both options by adding the following to `.swiftformat`:
```
--disable hoistTry
--disable hoistAwait
```

You can read more about the options [here](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#hoistTry) and [here](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#hoistAwait).

PR #53, opened 2023-05-26
KafkaProducer: Backpressure for Ack'd messages

felixschlegel:
### Motivation

Our current implementation of polling `librdkafka` for new message acknowledgements in `KafkaProducer` did not support backpressure. Also, it made use of some `weak` references which we generally want to avoid due to performance reasons.

### Modifications

* create a new class called `KafkaBackPressurePollingSystem`
* implement backpressure for `KafkaProducer`'s poll-loop with `KafkaBackPressurePollingSystem`
* fix typo in `SwiftKafkaTests`

### TODO

- [x] write sensible tests for this backpressure mechanism and the state machine
- [x] how to handle dependency injection into our `KafkaBackPressurePollingSystem`, given that some of its dependencies are not `Sendable`
- [x] discuss if we want to expose a `Kafka.run()` method to the user or if we want to keep the poll task to ourselves (cc: @FranzBusch )

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,25 +16,28 @@ let producer = try await KafkaProducer(
     logger: .kafkaTest // Your logger here
 )
 
-let runTask = Task {
-    await producer.run()
+await withThrowingTaskGroup(of: Void.self) { group in
+
+    group.addTask {
+        try await producer.run()
+    }
+
+    group.addTask {
+        let messageID = try await producer.sendAsync(
+            KafkaProducerMessage(
+                topic: "topic-name",
+                value: "Hello, World!"
+            )
+        )
+
+        for await acknowledgement in producer.acknowledgements {
+            // Check if acknowledgement belongs to the sent message
+        }
+
+        // Required
+        await producer.shutdownGracefully()
+    }
 }
-
-let messageID = try await producer.sendAsync(
-    KafkaProducerMessage(
-        topic: "topic-name",
-        value: "Hello, World!"
-    )
-)
-
-for await acknowledgement in producer.acknowledgements {
-    // Check if acknowledgement belongs to the sent message
-}
-
-runTask.cancel()
-
-// Required
-await producer.shutdownGracefully()
 ```
 
 ### Consumer API
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ The `sendAsync(_:)` method of `KafkaProducer` returns a message-id that can late
 ```swift
 let config = KafkaProducerConfig(bootstrapServers: ["localhost:9092"])
 
-let (producer, acknowledgements) = try await KafkaProducer.newProducer(
+let producer = try await KafkaProducer(
     config: config,
     logger: .kafkaTest // Your logger here
 )
@@ -32,7 +32,7 @@ await withThrowingTaskGroup(of: Void.self) { group in
             )
         )
 
-        for await acknowledgement in acknowledgements {
+        for await acknowledgement in producer.acknowledgements {
             // Check if acknowledgement belongs to the sent message
         }
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ The `sendAsync(_:)` method of `KafkaProducer` returns a message-id that can late
 ```swift
 let config = KafkaProducerConfig(bootstrapServers: ["localhost:9092"])
 
-let producer = try await KafkaProducer(
+let (producer, acknowledgements) = try await KafkaProducer.newProducer(
     config: config,
     logger: .kafkaTest // Your logger here
 )
@@ -32,7 +32,7 @@ await withThrowingTaskGroup(of: Void.self) { group in
             )
         )
 
-        for await acknowledgement in producer.acknowledgements {
+        for await acknowledgement in acknowledgements {
             // Check if acknowledgement belongs to the sent message
         }
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ The `sendAsync(_:)` method of `KafkaProducer` returns a message-id that can late
 ```swift
 let config = KafkaProducerConfig(bootstrapServers: ["localhost:9092"])
 
-let (producer, acknowledgements) = try await KafkaProducer.newProducer(
+let (producer, acknowledgements) = try await KafkaProducer.newProducerWithAcknowledgements(
     config: config,
     logger: .kafkaTest // Your logger here
 )
--- a/README.md
+++ b/README.md
@@ -18,10 +18,12 @@ let producer = try await KafkaProducer(
 
 await withThrowingTaskGroup(of: Void.self) { group in
 
+    // Run Task
     group.addTask {
         try await producer.run()
     }
 
+    // Task receiving acknowledgements
     group.addTask {
         let messageID = try await producer.sendAsync(
             KafkaProducerMessage(
--- a/README.md
+++ b/README.md
@@ -16,6 +16,10 @@ let producer = try await KafkaProducer(
     logger: .kafkaTest // Your logger here
 )
 
+let runTask = Task {
+    await producer.run()
+}
+
 let messageID = try await producer.sendAsync(
     KafkaProducerMessage(
         topic: "topic-name",
@@ -26,6 +30,8 @@ let messageID = try await producer.sendAsync(
 for await acknowledgement in producer.acknowledgements {
     // Check if acknowledgement belongs to the sent message
 }
+
+runTask.cancel()
 
 // Required
 await producer.shutdownGracefully()
--- a/Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
+++ b/Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
@@ -230,11 +230,11 @@ extension KafkaBackPressurePollingSystem {
         ///
         /// - Parameter continuation: The continuation that will be resumed once we are allowed to produce again.
         /// After resuming the continuation, our poll loop will start running again.
-        mutating func suspendLoop(continuation: CheckedContinuation<Void, Never>) {
+        fileprivate mutating func suspendLoop(continuation: CheckedContinuation<Void, Never>) {
             switch self.state {
             case .finished:
                 return
-            case .stopProducing(let existingContinuation) where existingContinuation == nil:
+            case .stopProducing(let existingContinuation) where existingContinuation != nil:
                 fatalError("Created leaking continuation")
             case .initial, .producing, .stopProducing:
                 self.state = .stopProducing(continuation)
--- a/Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
+++ b/Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
@@ -136,8 +136,8 @@ final class KafkaBackPressurePollingSystem {
                 do {
                     try await Task.sleep(for: pollInterval)
                 } catch {
-                    self.stateMachineLock.withLockedValue { $0.shutDown() }
-                    return
+                    let command = self.stateMachineLock.withLockedValue { $0.shutDown() }
+                    self.handleStateMachineCommand(command)
                 }
             case .suspendPollLoop:
                 await withTaskCancellationHandler {
@@ -145,10 +145,12 @@ final class KafkaBackPressurePollingSystem {
                         self.stateMachineLock.withLockedValue { $0.suspendLoop(continuation: continuation) }
                     }
                 } onCancel: {
-                    self.stateMachineLock.withLockedValue { $0.shutDown() }
+                    let command = self.stateMachineLock.withLockedValue { $0.shutDown() }
+                    self.handleStateMachineCommand(command)
                 }
             case .shutdownPollLoop:
-                self.stateMachineLock.withLockedValue { $0.shutDown() }
+                let command = self.stateMachineLock.withLockedValue { $0.shutDown() }
+                self.handleStateMachineCommand(command)
                 return
             }
         }
@@ -162,42 +164,51 @@ final class KafkaBackPressurePollingSystem {
             return
         }
 
-        let messageID = UInt(bitPattern: messagePointer.pointee._private)
-
-        let messageResult: Element
-        do {
-            let message = try KafkaAcknowledgedMessage(messagePointer: messagePointer, id: messageID)
-            messageResult = .success(message)
-        } catch {
-            guard let error = error as? KafkaAcknowledgedMessageError else {
-                fatalError("Caught error that is not of type \(KafkaAcknowledgedMessageError.self)")
-            }
-            messageResult = .failure(error)
-        }
-
-        self.stateMachineLock.withLockedValue { stateMachine in
-            let yieldResult = stateMachine.yield(messageResult)
+        let command = self.stateMachineLock.withLockedValue { stateMachine in
+            let yieldResult = stateMachine.sequenceSource?.yield(messageResult)
             switch yieldResult {
             case .produceMore:
-                stateMachine.produceMore()
+                return stateMachine.produceMore()
             case .stopProducing:
                 stateMachine.stopProducing()
-            case .dropped:
-                stateMachine.shutDown()
+                return nil
+            case .dropped, .none:
+                return stateMachine.shutDown()
             }
         }
+        self.handleStateMachineCommand(command)
+
         // The messagePointer is automatically destroyed by librdkafka
         // For safety reasons, we only use it inside of this closure
+    }
+
+    /// Handles an optional command that the ``KafkaBackPressurePollingSystem/StateMachine`` has wants us to run.
+    ///
+    /// - Parameter command: The command the ``KafkaBackPressurePollingSystem/StateMachine`` wants us to run.
+    private func handleStateMachineCommand(_ command: StateMachine.Command?) {
+        switch command {
+        case .resume(let continuation):
+            continuation?.resume()
+        case .finishSequenceSource:
+            self.stateMachineLock.withLockedValue { $0.sequenceSource?.finish() }
+        case .finishSequenceSourceAndResume(let continuation):
+            self.stateMachineLock.withLockedValue { $0.sequenceSource?.finish() }
+            continuation?.resume()
+        case .none:
+            break
+        }
     }
 }
 
 extension KafkaBackPressurePollingSystem: NIOAsyncSequenceProducerDelegate {
     func produceMore() {
-        self.stateMachineLock.withLockedValue { $0.produceMore() }
+        let command = self.stateMachineLock.withLockedValue { $0.produceMore() }
+        self.handleStateMachineCommand(command)
     }
 
     func didTerminate() {
-        self.stateMachineLock.withLockedValue { $0.shutDown() }
+        let command = self.stateMachineLock.withLockedValue { $0.shutDown() }
+        self.handleStateMachineCommand(command)
     }
 }
 
@@ -252,24 +263,36 @@ extension KafkaBackPressurePollingSystem {
             }
         }
 
+        /// Represents the commands that can be returned by a state machine
+        /// and shall be executed by the ``KafkaBackPressurePollingSystem``.
+        enum Command {
+            /// Resume the given continuation.
+            case resume(CheckedContinuation<Void, Never>?)
+            /// Invoke `.finish()` on the ``NIOAsyncSequence.Source``.
+            case finishSequenceSource
+            /// Resume the given continuation and invoke `.finish()` on the ``NIOAsyncSequence.Source``.
+            case finishSequenceSourceAndResume(CheckedContinuation<Void, Never>?)
+        }
+
         /// Our downstream consumer allowed us to produce more elements.
-        mutating func produceMore() {
+        mutating func produceMore() -> Command? {
             switch self.state {
             case .finished, .producing:
-                return
+                break
             case .stopProducing(let continuation):
-                continuation?.resume()
                 self.state = .producing
+                return .resume(continuation)
             case .initial:
                 self.state = .producing
             }
+            return nil
         }
 
         /// Our downstream consumer asked us to stop producing new elements.
         mutating func stopProducing() {
             switch self.state {
             case .finished, .stopProducing:
-                return
+                break
             case .initial:
                 fatalError("\(#function) is not supported in state \(self.state)")
             case .producing:
@@ -293,28 +316,17 @@ extension KafkaBackPressurePollingSystem {
         }
 
         /// Shut down the state machine and finish producing elements.
-        mutating func shutDown() {
-            self.sequenceSource?.finish()
-
+        mutating func shutDown() -> Command? {
             switch self.state {
             case .finished:
-                return
+                return nil
             case .initial, .producing:
-                self.sequenceSource?.finish()
                 self.state = .finished
+                return .finishSequenceSource
             case .stopProducing(let continuation):
-                continuation?.resume()
-                self.sequenceSource?.finish()
                 self.state = .finished
+                return .finishSequenceSourceAndResume(continuation)
             }
-        }
-
-        /// Yields a new elements to the ``NIOAsyncSequenceProducer``.
-        fileprivate func yield(_ element: Element) -> Producer.Source.YieldResult {
-            guard let source = self.sequenceSource else {
-                fatalError("\(#function) requires the sequenceSource variable to be non-nil")
-            }
-            return source.yield(element)
         }
     }
 }
--- a/Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
+++ b/Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
@@ -122,7 +122,12 @@ final class KafkaBackPressurePollingSystem {
     /// - Parameter pollInterval: The desired time interval between two consecutive polls.
     /// - Returns: An awaitable task representing the execution of the poll loop.
     func run(pollInterval: Duration) async {
-        actionLoop: while true {
+        let state = self.stateMachineLock.withLockedValue { $0.state }
+        guard case .initial = state else {
+            fatalError("Poll loop must not be started more than once")
+        }
+
+        while true {
             let action = self.stateMachineLock.withLockedValue { $0.nextPollLoopAction() }
 
             switch action {
@@ -132,7 +137,7 @@ final class KafkaBackPressurePollingSystem {
                     try await Task.sleep(for: pollInterval)
                 } catch {
                     self.stateMachineLock.withLockedValue { $0.shutDown() }
-                    break actionLoop
+                    return
                 }
             case .suspendPollLoop:
                 await withTaskCancellationHandler {
@@ -144,13 +149,14 @@ final class KafkaBackPressurePollingSystem {
                 }
             case .shutdownPollLoop:
                 self.stateMachineLock.withLockedValue { $0.shutDown() }
-                break actionLoop
+                return
             }
         }
     }
 
     /// The delivery report callback function that handles acknowledged messages.
     private(set) lazy var deliveryReportCallback: (UnsafePointer<rd_kafka_message_t>?) -> Void = { messagePointer in
+        // TODO: message pointer to Acknowledged message conversion at a lower level (e.g. RDKafkaConfig)
         guard let messagePointer = messagePointer else {
             self.logger.error("Could not resolve acknowledged message")
             return
@@ -203,21 +209,6 @@ extension KafkaBackPressurePollingSystem {
         /// The ``NIOAsyncSequenceProducer.Source`` used for yielding the messages to the ``NIOAsyncSequenceProducer``.
         var sequenceSource: Producer.Source? // TODO: make sendable
 
-        /// The events that can be triggered by the ``KafkaBackPressurePollingSystem``.
-        enum Event {
-            /// Produce more elements.
-            case produceMore
-            /// Suspend the poll loop until the continuation is resumed.
-            case suspendLoop(CheckedContinuation<Void, Never>)
-            /// Our `Task` was cancelled why we were waiting for the loop to be unsuspended.
-            case loopSuspensionCancelled
-            /// Request to stop producing elements. When this reaches the poll loop it will escalate
-            /// to a full suspension.
-            case stopProducing
-            /// Close the entire ``KafkaBackPressurePollingSystem`` with its poll loop.
-            case shutdown
-        }
-
         /// The possible states of the state machine.
         enum State {
             /// Initial state.
@@ -268,7 +259,7 @@ extension KafkaBackPressurePollingSystem {
                 return
             case .stopProducing(let continuation):
                 continuation?.resume()
-                fallthrough
+                self.state = .producing
             case .initial:
                 self.state = .producing
             }
@@ -279,10 +270,10 @@ extension KafkaBackPressurePollingSystem {
             switch self.state {
             case .finished, .stopProducing:
                 return
+            case .initial:
+                fatalError("\(#function) is not supported in state \(self.state)")
             case .producing:
                 self.state = .stopProducing(nil)
-            default:
-                fatalError("\(#function) is not supported in state \(self.state)")
             }
         }
 
@@ -294,8 +285,8 @@ extension KafkaBackPressurePollingSystem {
             switch self.state {
             case .finished:
                 return
-            case .stopProducing(let existingContinuation) where existingContinuation != nil:
-                fatalError("Created leaking continuation")
+            case .stopProducing(.some):
+                fatalError("Internal state inconsistency. Run loop is running more than once")
             case .initial, .producing, .stopProducing:
                 self.state = .stopProducing(continuation)
             }
@@ -308,10 +299,11 @@ extension KafkaBackPressurePollingSystem {
             switch self.state {
             case .finished:
                 return
+            case .initial, .producing:
+                self.sequenceSource?.finish()
+                self.state = .finished
             case .stopProducing(let continuation):
                 continuation?.resume()
-                fallthrough
-            default:
                 self.sequenceSource?.finish()
                 self.state = .finished
             }
--- a/Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
+++ b/Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
@@ -157,9 +157,8 @@ final class KafkaBackPressurePollingSystem {
     }
 
     /// The delivery report callback function that handles acknowledged messages.
-    private(set) lazy var deliveryReportCallback: (UnsafePointer<rd_kafka_message_t>?) -> Void = { messagePointer in
-        // TODO: message pointer to Acknowledged message conversion at a lower level (e.g. RDKafkaConfig)
-        guard let messagePointer = messagePointer else {
+    private(set) lazy var deliveryReportCallback: (RDKafkaConfig.KafkaAcknowledgementResult?) -> Void = { messageResult in
+        guard let messageResult else {
             self.logger.error("Could not resolve acknowledged message")
             return
         }
@@ -215,6 +214,7 @@ extension KafkaBackPressurePollingSystem: NIOAsyncSequenceProducerDelegate {
 extension KafkaBackPressurePollingSystem {
     /// The state machine used by the ``KafkaBackPressurePollingSystem``.
     struct StateMachine: Sendable {
+        // TODO: these are not handled optimally
         /// Closure that takes care of polling `librdkafka` for new messages.
         var pollClosure: (() -> ())?
         /// The ``NIOAsyncSequenceProducer.Source`` used for yielding the messages to the ``NIOAsyncSequenceProducer``.
--- a/Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
+++ b/Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
@@ -1,0 +1,276 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the swift-kafka-gsoc open source project
+//
+// Copyright (c) 2022 Apple Inc. and the swift-kafka-gsoc project authors
+// Licensed under Apache License v2.0
+//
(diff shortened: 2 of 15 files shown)


Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,274 @@
...
+    typealias Element = Result<KafkaAcknowledgedMessage, KafkaAcknowledgedMessageError>
+    /// The producer type used in the system.
+    typealias Producer = NIOAsyncSequenceProducer<Element, NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark, KafkaBackPressurePollingSystem>
+
+    /// The state machine that manages the system's state transitions.
+    let stateMachine: StateMachine
+
+    // TODO: wrapped values are non-sendable
+    private let _client: NIOLockedValueBox<KafkaClient?>
+    private let _sequenceSource: NIOLockedValueBox<Producer.Source?>
```
FranzBusch, line 35: Rather than having state outside of the state machine just put it into the associated values of the state. This way they are protected by the lock. If an action has to do something with those just return it in the action.

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,274 @@
...
+    /// Runs the poll loop with the specified poll interval.
+    ///
+    /// - Parameter pollIntervalNanos: The poll interval in nanoseconds.
+    /// - Returns: An awaitable task representing the execution of the poll loop.
+    func run(pollIntervalNanos: UInt64) async {
+        actionLoop: while true {
+            let action = self.stateMachine.nextPollLoopAction()
+
+            switch action {
+            case .poll:
```
FranzBusch, line 87: Can we name this action `poolAndSleep`

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,274 @@
...
+                self.client?.withKafkaHandlePointer { handle in
+                    rd_kafka_poll(handle, 0)
+                }
+                try? await Task.sleep(nanoseconds: pollIntervalNanos)
```
FranzBusch, line 91: We shouldn't silence the error here. If our task got cancelled we should exit out of the run loop. Importantly though we probably want to finish the source then.

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,274 @@
...
+            /// The system up and producing acknowledgement messages.
+            case producing
+            /// We were asked to stop producing messages. However we still wait for the
+            /// poll loop to be suspended.
+            case suspended
+            /// The pool loop is currently suspended and we are waiting for an invocation
+            /// of `produceMore()` to continue producing messages.
+            case loopSuspended(CheckedContinuation<Void, Never>)
+            /// The system is shut down.
+            case shutDown
```
FranzBusch, line 199: Can we rename this to `finished` to follow our other state machine naming

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,274 @@
...
+    }
+}
+
+extension KafkaBackPressurePollingSystem {
+    // TODO: test state machine
+
+    /// The state machine used by the ``KafkaBackPressurePollingSystem``.
+    struct StateMachine {
+        /// The possible actions for the poll loop.
+        enum PollLoopAction {
```
FranzBusch, line 162: Can we move the actions right above the method that returns them. This makes it easier to understand what is going on.

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,274 @@
...
+                case .shutDown:
+                    return .shutdownPollLoop
+                }
+            }
+        }
+
+        /// Triggers an event in the state machine, causing a state transition.
+        ///
+        /// - Parameter event: The event to be triggered.
+        func eventTriggered(_ event: Event) {
```
FranzBusch, line 229: Event triggered is way too broad. Please introduce separate methods for all the various events you model here e.g. `pollSuspended`, `pollCancelled` etc.

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,274 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+import Logging
+import NIOConcurrencyHelpers
+import NIOCore
+
+// TODO: write test for evaluating right state transitions
```
FranzBusch, line 20: To make this testable instead of passing in a `KafkaClient` you can pass in a closure that polls. In real code you can pass in a closure that does the actual polling and in the test you can pass something in which you can assert on

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,274 @@
...
+
+        self._client = NIOLockedValueBox(nil)
+        self._sequenceSource = NIOLockedValueBox(nil)
+    }
+
+    /// Runs the poll loop with the specified poll interval.
+    ///
+    /// - Parameter pollIntervalNanos: The poll interval in nanoseconds.
+    /// - Returns: An awaitable task representing the execution of the poll loop.
+    func run(pollIntervalNanos: UInt64) async {
```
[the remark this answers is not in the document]
felixschlegel: Yeah works, will have to update our minimum Package version to macOS `.v13` / iOS `.v16` though 👍

felixschlegel:
Thanks for your review! I have implemented your changes and added 3 test cases 😄 
I still get warnings about conforming to `Sendable`. Please see my comments inline!

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,263 @@
...
+    func didTerminate() {
+        self.stateMachineLock.withLockedValue { $0.shutDown() }
+    }
+}
+
+extension KafkaBackPressurePollingSystem {
+    /// The state machine used by the ``KafkaBackPressurePollingSystem``.
+    struct StateMachine: Sendable {
+        /// The ``NIOAsyncSequenceProducer.Source`` used for yielding the messages to the ``NIOAsyncSequenceProducer``.
+        var sequenceSource: Producer.Source? // TODO: make sendable?
```
felixschlegel, line 139: [`NIOAsyncSequenceProducer.Source`](https://github.com/apple/swift-nio/blob/6213ba7a06febe8fef60563a4a7d26a4085783cf/Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift#L211) is not `Sendable` at the moment which is giving us a warning here. Should I try making it `Sendable` in `swift-nio` or do we have a workaround for that in our case here?

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,263 @@
...
+    /// The element type for the system, representing either a successful ``KafkaAcknowledgedMessage`` or a ``KafkaAcknowledgedMessageError``.
+    typealias Element = Result<KafkaAcknowledgedMessage, KafkaAcknowledgedMessageError>
+    /// The producer type used in the system.
+    typealias Producer = NIOAsyncSequenceProducer<Element, NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark, KafkaBackPressurePollingSystem>
+
+    /// The state machine that manages the system's state transitions.
+    let stateMachineLock: NIOLockedValueBox<StateMachine>
+
+    /// Closure that takes care of polling `librdkafka` for new messages.
+    var pollClosure: (() -> Void)?
```
felixschlegel, line 31: Atm this is a `var` as it captures `self` in `KafkaProducer` so it can get set _after_ our `KafkaProducer` has been initialized. However, having a `var` here makes us not conform to `Sendable`.

FranzBusch: We should move this closure into the state machine. Furthermore, I don't see why this should capture the `KafkaProducer` the only thing that we need is capturing the `KafkaClient` to call `poll` on it

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,263 @@
...
+    /// The ``NIOAsyncSequenceProducer.Source`` used for yielding the messages to the ``NIOAsyncSequenceProducer``.
+    var sequenceSource: Producer.Source? {
```
FranzBusch, line 33: Can't we pass this in the init or rather could we actually create the sequence in here and return it from a factory method?

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,263 @@
...
+        self.stateMachineLock = NIOLockedValueBox(StateMachine())
+        self.logger = logger
+    }
+
+    /// Runs the poll loop with the specified poll interval.
+    ///
+    /// - Parameter pollInterval: The desired time interval between two consecutive polls.
+    /// - Returns: An awaitable task representing the execution of the poll loop.
+    func run(pollInterval: Duration) async {
+        actionLoop: while true {
```
FranzBusch, line 62: We don't need the label here instead we can just use `return` instead of `break` further down

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,263 @@
...
+            let action = self.stateMachineLock.withLockedValue { $0.nextPollLoopAction() }
+
+            switch action {
+            case .pollAndSleep:
+                self.pollClosure?()
+                do {
+                    try await Task.sleep(for: pollInterval)
+                } catch {
+                    self.stateMachineLock.withLockedValue { $0.shutDown() }
```
FranzBusch, line 71: We probably have to call `source.finish()` here?

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,263 @@
...
+                }
+            case .shutdownPollLoop:
+                self.stateMachineLock.withLockedValue { $0.shutDown() }
+                break actionLoop
+            }
+        }
+    }
+
+    /// The delivery report callback function that handles acknowledged messages.
+    private(set) lazy var deliveryReportCallback: (UnsafePointer<rd_kafka_message_t>?) -> Void = { messagePointer in
```
FranzBusch, line 90: In the scope of this PR can we make a change so that we don't use `UnsafePointer` here but rather bridge the call-back to something well typed inside the method where we set the callback?

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,263 @@
...
+            }
+        }
+
+        /// Our downstream consumer allowed us to produce more elements.
+        mutating func produceMore() {
+            switch self.state {
+            case .finished, .producing:
+                return
+            case .stopProducing(let continuation):
+                continuation?.resume()
```
FranzBusch, line 205: We shouldn't resume continuations inside the state machine. This should be an action and resumed outside of the sate machine. In general, the state machine should not execute any side effects rather just do plain state transitions and return actions

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
FranzBusch: The `StateMachine` doesn't necessarily need to be `Sendable` we are holding it in a `Lock` which makes it `Sendable` We just need to use `@unchecked Sendable` on the `KafkaBackPressurePollingSystem`

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,263 @@
...
+}
+
+extension KafkaBackPressurePollingSystem {
+    /// The state machine used by the ``KafkaBackPressurePollingSystem``.
+    struct StateMachine: Sendable {
+        /// The ``NIOAsyncSequenceProducer.Source`` used for yielding the messages to the ``NIOAsyncSequenceProducer``.
+        var sequenceSource: Producer.Source? // TODO: make sendable?
+
+        /// The events that can be triggered by the ``KafkaBackPressurePollingSystem``.
+        enum Event {
```
FranzBusch, line 142: Can we get rid of this now?

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        /// Our downstream consumer allowed us to produce more elements.
+        mutating func produceMore() {
+            switch self.state {
+            case .finished, .producing:
+                return
+            case .stopProducing(let continuation):
+                continuation?.resume()
+                fallthrough
```
FranzBusch, line 206: Please avoid using `fallthrough` it makes the code harder to reason about. Rather prefer explicit state transition in every case.

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,263 @@
...
+
+        /// Suspend the poll loop.
+        ///
+        /// - Parameter continuation: The continuation that will be resumed once we are allowed to produce again.
+        /// After resuming the continuation, our poll loop will start running again.
+        fileprivate mutating func suspendLoop(continuation: CheckedContinuation<Void, Never>) {
+            switch self.state {
+            case .finished:
+                return
+            case .stopProducing(let existingContinuation) where existingContinuation != nil:
```
FranzBusch, line 232: NIT
```suggestion
            case .stopProducing(.some):
```

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,263 @@
...
+                fatalError("Created leaking continuation")
```
FranzBusch, line 233:
```suggestion
                fatalError("Internal state inconsistency. Run loop is running more than once")
```

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
```diff
@@ -0,0 +1,263 @@
...
+                continuation?.resume()
+                fallthrough
+            default:
+                self.sequenceSource?.finish()
+                self.state = .finished
+            }
+        }
+
+        /// Yields a new elements to the ``NIOAsyncSequenceProducer``.
+        fileprivate func yield(_ element: Element) -> Producer.Source.YieldResult {
```
FranzBusch, line 256: Please yield outside of the state machine

FranzBusch (changes requested):
We are getting close! The state machine looks already quite good we just have to move some of the things outside of the state machine. I left some comments inline

Sources/SwiftKafka/KafkaBackPressurePollingSystem.swift
felixschlegel: `source.finish()` gets invoked by `shutDown()` :+1:

FranzBusch: Yes but that's inside the state machine and we want to have this outside the state machine! Since the state machine should not execute any side effects

felixschlegel:
@FranzBusch I have made the `KafkaBackPressurePollingSystem` generic in https://github.com/swift-server/swift-kafka-gsoc/pull/53/commits/624477c7a39c98e717f6be65ec712d734dbc6556, let me know what you think!

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,299 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+import Logging
+import NIOConcurrencyHelpers
+import NIOCore
+
+/// Our `AsyncSequence` implementation wrapping an `NIOAsyncSequenceProducer`.
+public struct KafkaAsyncSequence<Element, BackPressure: NIOAsyncSequenceProducerBackPressureStrategy>: AsyncSequence {
```
FranzBusch, line 21: Let's not make the backpressure strategy generic but rather choose high/low for the user here

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,299 @@
...
+    typealias PollingSystem = KafkaPollingSystem<Element, BackPressure>
+    typealias WrappedSequence = NIOAsyncSequenceProducer<Element, BackPressure, PollingSystem>
+    let wrappedSequence: WrappedSequence
+
+    /// Our `AsynceIteratorProtocol` implementation wrapping `NIOAsyncSequenceProducer.AsyncIterator`.
+    public struct KafkaAsyncIterator: AsyncIteratorProtocol {
+        let wrappedIterator: NIOAsyncSequenceProducer<Element, BackPressure, PollingSystem>.AsyncIterator
+
+        public mutating func next() async -> Element? {
```
FranzBusch, line 30: This probably should throw as well. From my understanding of the librd docs there are cases where the polling will throw an error

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,299 @@
...
+/// A back-pressure aware polling system for managing the poll loop that polls `librdkafka` for new acknowledgements.
+final class KafkaPollingSystem<
+    Element,
+    BackPressure: NIOAsyncSequenceProducerBackPressureStrategy
+>: Sendable {
+    /// The producer type used in the system.
+    typealias Producer = NIOAsyncSequenceProducer<Element, BackPressure, KafkaPollingSystem>
+
+    /// The state machine that manages the system's state transitions.
+    let stateMachineLock: NIOLockedValueBox<StateMachine>
```
FranzBusch, line 49: Can we make this private and rename it to `stateMachine`?

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,299 @@
...
+
+    /// Closure that takes care of polling `librdkafka` for new messages.
+    var pollClosure: (() -> Void)? {
```
FranzBusch, line 52: Why is this a var again? Can't we get this in the factory method?

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,299 @@
...
+            }
+        }
+    }
+
+    /// Initializes the ``KafkaBackPressurePollingSystem``.
+    /// Private initializer. The ``KafkaBackPressurePollingSystem`` is not supposed to be initialized directly.
+    /// It must rather be initialized using the ``KafkaBackPressurePollingSystem.createSystemAndSequence`` function.
+    /* USE createSystemAndSequence, TESTING only */ init() {
+        self.stateMachineLock = NIOLockedValueBox(StateMachine())
+    }
```
FranzBusch, line 70: Why do we need this for testing and can't use the factory method?

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,299 @@
...
+        // terminate the underlying source.
+        let acknowledgementsSourceAndSequence = NIOAsyncSequenceProducer.makeSequence(
+            elementType: Element.self,
+            backPressureStrategy: backPressureStrategy,
+            delegate: pollingSystem
+        )
+
+        pollingSystem.stateMachineLock.withLockedValue { stateMachine in
+            stateMachine.sequenceSource = acknowledgementsSourceAndSequence.source
+        }
```
FranzBusch, line 95: Can we rather pass this in the init of the `KafkaPollingSystem`?

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,299 @@
...
+    /// Runs the poll loop with the specified poll interval.
+    ///
+    /// - Parameter pollInterval: The desired time interval between two consecutive polls.
+    /// - Returns: An awaitable task representing the execution of the poll loop.
+    func run(pollInterval: Duration) async {
+        self.stateMachineLock.withLockedValue { stateMachine in
+            guard stateMachine.running == false else {
+                fatalError("Poll loop must not be started more than once")
+            }
+            stateMachine.running = true
```
FranzBusch, line 113: This should be done inside a method of the state machine. Let's add a method `run()` to the state machine as well that returns a `RunAction`.

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,299 @@
...
+
+            switch action {
+            case .pollAndSleep:
+                // Poll Kafka for new acknowledgements and sleep for the given
+                // pollInterval to avoid hot looping.
+                self.pollClosure?()
+                do {
+                    try await Task.sleep(for: pollInterval)
+                } catch {
+                    let command = self.stateMachineLock.withLockedValue { $0.shutDown() }
```
FranzBusch, line 127: We have to re throw the error here

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,299 @@
...
+
+extension KafkaPollingSystem {
+    /// The state machine used by the ``KafkaBackPressurePollingSystem``.
+    struct StateMachine {
+        /// A flag that determines if the ``run()`` method has already been invoked.
+        var running = false
+        /// Closure that takes care of polling `librdkafka` for new messages.
+        var pollClosure: (() -> Void)?
+        /// The ``NIOAsyncSequenceProducer.Source`` used for yielding the messages to the ``NIOAsyncSequenceProducer``.
+        var sequenceSource: Producer.Source?
```
FranzBusch, line 188: Let's move all the state into the associated values of the cases

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,299 @@
...
+            case .suspendPollLoop:
+                // The downstream consumer asked us to stop sending new messages.
+                // We therefore await until we are unsuspended again.
+                await withTaskCancellationHandler {
+                    await withCheckedContinuation { continuation in
+                        self.stateMachineLock.withLockedValue { $0.suspendLoop(continuation: continuation) }
+                    }
+                } onCancel: {
+                    let command = self.stateMachineLock.withLockedValue { $0.shutDown() }
+                    self.handleStateMachineCommand(command)
```
FranzBusch, line 139: We have to resume the continuation with the `CancellationError` here

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,299 @@
...
+                // Trigger the poll loop to suspend.
+                return .suspendPollLoop
+            case .finished:
+                return .shutdownPollLoop
+            }
+        }
+
+        /// Represents the commands that can be returned by a state machine
+        /// and shall be executed by the ``KafkaBackPressurePollingSystem``.
+        enum Command {
```
FranzBusch, line 235: We can please stick to the `Action` naming and introduce separate actions for each method

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,299 @@
...
+            case .producing:
+                self.state = .stopProducing(nil)
+            }
+        }
+
+        /// Suspend the poll loop.
+        ///
+        /// - Parameter continuation: The continuation that will be resumed once we are allowed to produce again.
+        /// After resuming the continuation, our poll loop will start running again.
+        fileprivate mutating func suspendLoop(continuation: CheckedContinuation<Void, Never>) {
```
FranzBusch, line 274: Why is this `fileprivate`?

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,299 @@
...
+                return
+            case .stopProducing(.some):
+                fatalError("Internal state inconsistency. Run loop is running more than once")
+            case .initial, .producing, .stopProducing:
+                self.state = .stopProducing(continuation)
+            }
+        }
+
+        /// Shut down the state machine and finish producing elements.
+        mutating func shutDown() -> Command? {
```
FranzBusch, line 286:
```suggestion
        mutating func terminate() -> Command? {
```

FranzBusch (changes requested):
Getting close! Some more comments

Sources/SwiftKafka/KafkaPollingSystem.swift
felixschlegel: We cannot pass everything on `init` as we have a lot of circular dependencies, e.g. `NIOAsyncSequenceProducer` needs `KafkaPollingSystem` as its delegate, `KafkaPollingSystem` needs `NIOAsyncSequenceProducer.Source` to produce new elements. So at some point, we have to assign the references after initialization.

Sources/SwiftKafka/KafkaPollingSystem.swift
felixschlegel: From [the documentation](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#ad50c431e3a29d14da534db49bd0682a4), `rd_kafka_poll` should not be an issue. However, [`rd_kafka_consumer_poll`](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#a65cc6cb9bd72c4084f074af0361ceddf) can result in errors. But I think it's better if we do that in the PR dedicated for `KafkaConsumer`, right?

README.md
```diff
@@ -27,6 +31,8 @@ for await acknowledgement in producer.acknowledgements {
     // Check if acknowledgement belongs to the sent message
 }
 
+runTask.cancel()
+
 // Required
 await producer.shutdownGracefully()
```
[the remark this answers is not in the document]
felixschlegel: #57

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,430 @@
...
+                let action = self.stateMachine.withLockedValue { $0.terminate() }
+                self.handleTerminateAction(action)
+            }
+        }
+    }
+
+    /// Yield new elements to the underlying `NIOAsyncSequenceProducer`.
+    func yield(_ element: Element) {
+        self.stateMachine.withLockedValue { stateMachine in
+            switch stateMachine.state {
```
FranzBusch, line 109: Why are we switching on the state here? We should switch on the state inside the state machine and return an action

Tests/IntegrationTests/SwiftKafkaTests.swift
```diff
@@ -70,6 +70,9 @@ final class SwiftKafkaTests: XCTestCase {
 
     func testProduceAndConsumeWithConsumerGroup() async throws {
         let producer = try await KafkaProducer(config: producerConfig, logger: .kafkaTest)
+        let _ = Task {
+            try await producer.run()
+        }
```
FranzBusch, line 75: Let's use task groups here instead of unstructured tasks

README.md
```diff
@@ -16,19 +16,28 @@ let producer = try await KafkaProducer(
...
 
-for await acknowledgement in producer.acknowledgements {
-    // Check if acknowledgement belongs to the sent message
-}
+    group.addTask {
+        try await producer.run()
+    }
+
+    group.addTask {
+        let messageID = try await producer.sendAsync(
```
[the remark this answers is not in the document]
felixschlegel: #59

Tests/SwiftKafkaTests/KafkaProducerTests.swift
```diff
@@ -54,120 +54,172 @@ final class KafkaProducerTests: XCTestCase {
...
+                )
+
+                do {
+                    try await producer.sendAsync(message)
+                    XCTFail("Method should have thrown error")
+                } catch {}
+
+                // This subscribes to the acknowledgements stream and immediately terminates the stream.
+                // Required to kill the run task.
+                var iterator: KafkaAsyncSequence<KafkaProducer.Acknowledgement>.AsyncIterator? = producer
```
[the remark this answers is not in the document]
FranzBusch: Since we dropping the iterator here the `didTerminate` callback on the delegate should be triggered which should lead to the `run()` task to terminate. Can we check that the `didTerminate` is triggered?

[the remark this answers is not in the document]
felixschlegel: The solution I came up with now is quite simple: I am just invoking `self.pollingSystem.terminate()` inside of `KafkaProducer.shutDownGracefully`. By doing that we don't have to worry about if the user subscribes to our `acknowledgements` `AsyncSequence` or not.

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+    /// Stop producing new elements to the
+    /// `source` ``NIOAsyncSequenceProducer``.
+    private func stopProducing() {
+        self.stateMachine.withLockedValue { $0.stopProducing() }
+    }
+
+    /// Kill the ``KafkaPollingSystem``'s poll loop and free its resources.
+    func terminate(_ error: Error? = nil) {
```
FranzBusch, line 180: Can this be private?

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,459 @@
...
+            self.terminate()
+        case .none:
+            break
+        }
+    }
+
+    /// Determines the next action to be taken in the poll loop based on the current state.
+    ///
+    /// - Returns: The next action for the poll loop.
+    func nextPollLoopAction() -> KafkaPollingSystem.StateMachine.PollLoopAction {
```
FranzBusch, line 169: Do we really need a separate method for this?

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,459 @@
...
+        }
+
+        /// Actions to take after ``run()`` has been invoked on the ``KafkaPollingSystem/StateMachine``.
+        enum RunAction {
+            /// Do nothing, the poll loop is already terminated.
+            case alreadyClosed
+            /// The poll loop is already running. ``run()`` should not be invoked more than once!
+            case alreadyRunning
+            /// The poll loop can now be started.
+            case startLoop
```
FranzBusch, line 273: One last nit here w.r.t. naming. Can you make sure all actions are named in way that they tell what do to e.g. `throwError` like you named `startLoop`

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,459 @@
...
+        mutating func run() -> RunAction {
+            switch self.state {
+            case .uninitialized:
+                fatalError("\(#function) invoked while still in state .uninitialized")
+            case .idle(let source, let pollClosure, let running):
+                guard running == false else {
+                    return .alreadyRunning
+                }
+                self.state = .idle(source: source, pollClosure: pollClosure, running: true)
+            case .producing(let source, let pollClosure, let running):
```
FranzBusch, line 285: Can this happen? How can we be in `producing` before `run` was called?

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,459 @@
...
+            case .stopProducing(let source, let continuation, let pollClosure, let running):
+                guard running == false else {
+                    return .alreadyRunning
+                }
+                self.state = .stopProducing(source: source, continuation: continuation, pollClosure: pollClosure, running: true)
+            case .finished:
+                return .alreadyClosed
+            }
+
+            return .startLoop
```
FranzBusch, line 299: Please add a return inside every individual case. That makes it easier to reason about and you don't have to scan until the end of the method

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,459 @@
...
+        }
+
+        /// The possible actions for the poll loop.
+        enum PollLoopAction {
+            /// Ask `librdkakfa` to receive new message acknowledgements at a given poll interval.
+            case pollAndSleep(pollClosure: (() -> Void)?)
```
FranzBusch, line 305: This can't be optional right?

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,459 @@
...
+    struct StateMachine {
+        /// The possible states of the state machine.
+        enum State {
+            /// The state machine has been initialized with init() but is not yet Initialized
+            /// using `func initialize()` (required).
+            case uninitialized
+            /// Initialized state (idle).
+            case idle(
+                source: Producer.Source?,
+                pollClosure: (() -> Void)?,
```
FranzBusch, line 230: Why are the `source` and the `pollClosure` optional here?

Sources/SwiftKafka/KafkaPollingSystem.swift
```diff
@@ -0,0 +1,459 @@
...
+
+        /// Yield new elements to the underlying `NIOAsyncSequenceProducer`.
+        mutating func yield(_ element: Element) -> YieldAction? {
+            switch self.state {
+            case .uninitialized:
+                fatalError("\(#function) invoked while still in state .uninitialized")
+            case .idle(let source, _, _), .producing(let source, _, _), .stopProducing(let source, _, _, _):
+                // We can also yield when in .stopProducing,
+                // the AsyncSequenceProducer will buffer for us
+                let yieldResult = source?.yield(element)
```
FranzBusch, line 349: Please don't yield inside the state machine. Yielding is a side effect so please return that as an action.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -16,43 +16,13 @@ import Crdkafka
...
-        return AcknowledgedMessagesAsyncIterator(wrappedIterator: self.wrappedSequence.makeAsyncIterator())
-    }
-}
-
 /// Send messages to the Kafka cluster.
 /// Please make sure to explicitly call ``shutdownGracefully(timeout:)`` when the ``KafkaProducer`` is not used anymore.
 /// - Note: When messages get published to a non-existent topic, a new topic is created using the ``KafkaTopicConfig``
 /// configuration object (only works if server has `auto.create.topics.enable` property set).
 public actor KafkaProducer {
+    public typealias Acknowledgement = Result<KafkaAcknowledgedMessage, KafkaAcknowledgedMessageError>
```
FranzBusch, line 24: Why are we typealiasing this here? Let's not introduce this please but spell it out

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -162,14 +129,25 @@ public actor KafkaProducer {
             }
         }
 
+        // Kill poll loop in polling system
+        self.pollingSystem.terminate()
```
FranzBusch, line 133: I am not sure that this is enough. We should figure out what happens in a system where we don't poll anymore but still produce e.g. nobody consumes the acknowledgements sequence and it gets back pressured. Will rdkafka buffer then and at some point also back pressure or will we just OOM?

Sources/SwiftKafka/KafkaPollingSystem.swift
felixschlegel: I thought this might be possible e.g. when `produceMore()` gets called upon the `NIOAsyncSequenceProducerDelegate` before `run()` gets invoked

FranzBusch: That won't happen. `produceMore()` only gets called after `yield()` returns `stopProducing`

Sources/SwiftKafka/KafkaPollingSystem.swift
felixschlegel: Not really, but given that our `stateMachine` is `private` this provides a safe way to access the poll loop's state for testing. There are not too many things we can assert on here in testing. Alternatively we can make `KafkaPollingSystem.yield(_:)` return the next action but this would also then just serve the purpose on being able to assert on the `KafkaPollingSystem`. In general all of this is quite hard as the `KafkaPollingSystem` does not expose that much information to the outside world.

Sources/SwiftKafka/KafkaPollingSystem.swift
felixschlegel: Yes this would be no problem because we can also invoke `didTerminate()` from outside, but I thought that the name `terminate()` made more sense, also because `didTerminate()` is a delegate method.

(My implementation needs an `internal` method for terminating the `KafkaPollingSystem` because I invoke this from `KafkaPollingSystem.shutDownGracefully`)

Sources/SwiftKafka/KafkaPollingSystem.swift
felixschlegel: You're right, this was from a previous implementation!

Sources/SwiftKafka/KafkaProducer.swift
felixschlegel: I believe [`rd_kafka_flush()`](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#aaff06c4372bce917c17f3c1a5d8b205d) got us covered here (we already use this).

felixschlegel:
Having back pressure makes no sense as `librdkafka` is queuing the messages regardless, closing, but keeping branch for reference

PR #54, opened 2023-06-02
`KafkaConsumerConfig` add `SwiftKafka` options

felixschlegel:
### Motivation

We want to use the `KafkaConsumerConfig` to determine if the `KafkaConsumer` is part of a consumer-group (subscription) or assigned to a single topic partition (assignment). This should also be implemented so that errors are caught at compile time. Furthermore, we want users to be able to determine a `BackPressureStrategy` for their respective `KafkaClient`.

### Modifications

* add two new configuration options to `KafkaConsumerConfig`: `ConsumptionStrategy` and `BackPressureStrategy`
* remove `KafkaConsumerConfig`'s `groupID` option. This is now taken care of by the `consumptionStrategy` option, which is better as this ensures at compile time that now groupID mismatch errors occur
* rename `KafkaProducerConfig` back to `ConfigEnums` as it is not really meant for shared configurations, but rather as a namespace for _all_ configuration options
* update README

### DIFF
--- a/Sources/SwiftKafka/Configuration/ConfigEnums.swift
+++ b/Sources/SwiftKafka/Configuration/ConfigEnums.swift
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 /// Collection of `enum` types used in the configuration structs this library provides.
-public enum KafkaSharedConfiguration {
+public enum ConfigEnums {
     /// Available debug contexts to enable.
     public struct DebugOption: Hashable, Equatable, CustomStringConvertible {
         public let description: String
--- a/Sources/SwiftKafka/Configuration/KafkaConsumerConfig.swift
+++ b/Sources/SwiftKafka/Configuration/KafkaConsumerConfig.swift
@@ -12,16 +12,26 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Crdkafka
+
 public struct KafkaConsumerConfig: Hashable, Equatable {
+    // MARK: - SwiftKafka-specific Config properties
+
+    /// The backpressure strategy to be used for receiveing message acknowledgements.
+    public var backPressureStrategy: ConfigEnums.BackPressureStrategy = .highLowWatermark(
+        lowWatermark: 10,
+        highWatermark: 50
+    )
+
+    /// The strategy used for consuming messages.
+    /// See ``ConfigEnums/ConsumptionStrategy`` for more information.
+    public var consumptionStrategy: ConfigEnums.ConsumptionStrategy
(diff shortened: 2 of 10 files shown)


Sources/SwiftKafka/Configuration/ConfigEnums.swift
```diff
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 /// Collection of `enum` types used in the configuration structs this library provides.
-public enum KafkaSharedConfiguration {
+public enum ConfigEnums {
     /// Available debug contexts to enable.
```
FranzBusch, line 17: This seems wrong

Sources/SwiftKafka/Configuration/KafkaConsumerConfig.swift
```diff
@@ -394,46 +431,107 @@ public struct KafkaConsumerConfig: Hashable, Equatable {
...
+        }
+
+        /// A back pressure strategy based on high and low watermarks.
+        ///
+        /// The consumer maintains a buffer size between a low watermark and a high watermark
+        /// to control the flow of incoming messages.
+        ///
+        /// - Parameter lowWatermark: The lower threshold for the buffer size.
+        /// - Parameter highWatermark: The upper threshold for the buffer size.
+        public static func highLowWatermark(lowWatermark: Int, highWatermark: Int) -> BackPressureStrategy {
```
FranzBusch, line 493: Let's rename this
```suggestion
        public static func watermark(low: Int, high: Int) -> BackPressureStrategy {
```

FranzBusch (changes requested):

Sources/SwiftKafka/Configuration/KafkaConsumerConfig.swift
```diff
@@ -431,9 +469,70 @@ public struct KafkaConsumerConfig: Hashable, Equatable {
...
+            self._internal = backPressureStrategy
+        }
+
+        /// A back pressure strategy based on high and low watermarks.
+        ///
+        /// The consumer maintains a buffer size between a low watermark and a high watermark
+        /// to control the flow of incoming messages.
+        ///
+        /// - Parameter lowWatermark: The lower threshold for the buffer size.
+        /// - Parameter highWatermark: The upper threshold for the buffer size.
```
FranzBusch, line 493: Can we update the parameter names

felixschlegel:
> Almost there just some doc comment nits

Arrgh! Fixed it ✅

FranzBusch (approved):

PR #55, opened 2023-06-05
KafkaProducerConfig: add `allow.auto.create.topics` option

felixschlegel:
### Motivation

* the `allow.auto.create.topics` was missing in our `KafkaProducerConfig`

> **Note**: the default value for `allow.auto.create.topics` is `true` for `KafkaProducerConfig` and `false` for `KafkaConsumerConfig` (this is adapted from `librdkafka`)

### Modifications

* add `allow.auto.create.topics` option to `KafkaProducerConfig`

### DIFF
--- a/Sources/SwiftKafka/Configuration/KafkaProducerConfig.swift
+++ b/Sources/SwiftKafka/Configuration/KafkaProducerConfig.swift
@@ -59,12 +59,9 @@ public struct KafkaProducerConfig: Hashable, Equatable {
         set { self.dictionary["message.send.max.retries"] = String(newValue) }
     }
 
-    /// Allow automatic topic creation on the broker when subscribing to or assigning non-existent topics.
+    /// Allow automatic topic creation on the broker when producing to non-existent topics.
     /// The broker must also be configured with auto.create.topics.enable=true for this configuration to take effect.
-    /// Note: the default value (true) for the producer is different from the default value (false) for the consumer.
-    /// Further, the consumer default value is different from the Java consumer (true), and this property
-    /// is not supported by the Java producer.
-    /// Requires broker version >= 0.11.0.0, for older broker versions only the broker configuration applies.
+    /// Default value: `true`
     public var allowAutoCreateTopics: Bool {
         get { self.dictionary.getBool("allow.auto.create.topics") ?? true }
         set { self.dictionary["allow.auto.create.topics"] = String(newValue) }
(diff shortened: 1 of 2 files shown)


Sources/SwiftKafka/Configuration/KafkaProducerConfig.swift
```diff
@@ -59,6 +59,12 @@ public struct KafkaProducerConfig: Hashable, Equatable {
         set { self.dictionary["message.send.max.retries"] = String(newValue) }
     }
 
+    /// Allow automatic topic creation on the broker when subscribing to or assigning non-existent topics. The broker must also be configured with auto.create.topics.enable=true for this configuration to take effect. Note: the default value (true) for the producer is different from the default value (false) for the consumer. Further, the consumer default value is different from the Java consumer (true), and this property is not supported by the Java producer. Requires broker version >= 0.11.0.0, for older broker versions only the broker configuration applies.
```
FranzBusch, line 62: Can we break this into multiple lines?

felixschlegel: Broke the documentation comment down into multiple lines ✅

felixschlegel: I do not see the issue with default values:

In `KafkaProducerConfig` it should be `true`,
in `KafkaConsumerConfig` it should be `false` by default

Sources/SwiftKafka/Configuration/KafkaProducerConfig.swift
```diff
@@ -59,6 +59,17 @@ public struct KafkaProducerConfig: Hashable, Equatable {
         set { self.dictionary["message.send.max.retries"] = String(newValue) }
     }
 
+    /// Allow automatic topic creation on the broker when subscribing to or assigning non-existent topics.
+    /// The broker must also be configured with auto.create.topics.enable=true for this configuration to take effect.
+    /// Note: the default value (true) for the producer is different from the default value (false) for the consumer.
+    /// Further, the consumer default value is different from the Java consumer (true), and this property
+    /// is not supported by the Java producer.
+    /// Requires broker version >= 0.11.0.0, for older broker versions only the broker configuration applies.
```
FranzBusch, line 67: The problem with the default value is that you are stating for the producer it is false here but if you query this property then it will evaluate to `true`. However at runtime if nothing is set RDKafka will use `false` for it.

FranzBusch (approved):

PR #66, opened 2023-06-19
`KafkaConsumer` Refactoring

felixschlegel:
### Motivation

Our `KafkaConsumer` should expose a `run()` method that serves the consumer queue and polls `librdkafka` for new messages and any other queued callbacks. This aligns with the new implementation of `KafkaProducer`.

##### Additionally

* get rid of `weak` reference to `self` in `KafkaConsumer`
* get rid of `DispatchQueue` in `KafkaConsumer`

### Modifications

* expose new method `KafkaConsumer.run()`
* write async wrapper method `KafkaClient.consumerPoll(timeout:)`
* replace `serialQueue` in `KafkaConsumer` with `StateMachine` that
  encapsulates all variables and can also be accessed from the
  `ConsumerMessagesAsyncSequence`
* close `KafkaConsumer` when `for await` loop of `AsyncSequence` is
  exited
* make `KafkaConsumer.shutdownGracefully` `public`
* remove `BackPressureStrategy` from `KafkaConsumerConfiguration`


Sources/SwiftKafka/KafkaClient.swift
```diff
@@ -56,6 +56,43 @@ final class KafkaClient {
...
+    /// Request a new message from the Kafka cluster.
+    ///
+    /// This method blocks for a maximum of `timeout` milliseconds.
+    ///
+    /// - Important: This method should only be invoked from ``KafkaConsumer``.
+    ///
+    /// - Parameter timeout: Maximum amount of milliseconds this method waits for a new message.
+    /// - Returns: A ``KafkaConsumerMessage`` or `nil` if there are no new messages.
+    /// - Throws: A ``KafkaError`` if the received message is an error message or malformed.
+    func consumerPoll(timeout: Int32) async throws -> KafkaConsumerMessage? {
```
FranzBusch, line 68: We should remove the `timeout` here otherwise people might actually block. Additionally, we should remove the `async` from this method. I don't see a reason why we are creating a continuation here. We should be able to just call `rd_kafka_consumer_poll` and bridge the message into a Swift type, destroy the pointer and return.

felixschlegel: Totally makes sense!

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -151,27 +160,34 @@ public final class KafkaConsumer {
...
+                    KafkaPartition.unassigned.rawValue
+                )
+            }
 
-        let result = self.client.withKafkaHandlePointer { handle in
-            rd_kafka_subscribe(handle, self.subscribedTopicsPointer)
-        }
+            let result = client.withKafkaHandlePointer { handle in
+                rd_kafka_subscribe(handle, subscribedTopicsPointer)
+            }
```
FranzBusch, line 186: For all these places where we use `withKafkaHandlePointer` can we just add a method on the `KafkaClient` instead? At best that method also already does the conversion to the correct error.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -185,102 +201,68 @@ public final class KafkaConsumer {
...
+            switch nextAction {
+            case .pollForAndYieldMessage(let pollInterval, let client, let source, let logger):
+                let messageResult: Result<KafkaConsumerMessage, KafkaError>
+                do {
+                    guard let message = try client.consumerPoll() else {
+                        break
+                    }
+                    messageResult = .success(message)
+                } catch let kafkaError as KafkaError {
+                    messageResult = .failure(kafkaError)
```
FranzBusch, line 243: If we fail polling shouldn't we shutdown everything?

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -185,102 +201,68 @@ public final class KafkaConsumer {
...
     /// - Parameter message: Last received message that shall be marked as read.
     /// - Throws: A ``KafkaError`` if committing failed.
     /// - Warning: This method fails if the `enable.auto.commit` configuration property is set to `true`.
     public func commitSync(_ message: KafkaConsumerMessage) async throws {
-        try await self.serializeWithThrowingContinuation { (continuation: CheckedContinuation<Void, Error>) in
-
+        try await withCheckedThrowingContinuation { continuation in
             do {
-                try self._commitSync(message)
+                try self._commitSync(message) // Blocks until commiting the offset is done
```
FranzBusch, line 265: Yeah that is problematic. We need to change that.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -289,72 +271,285 @@ public final class KafkaConsumer {
...
+            /// - Parameter client: Client used for handling the connection to the Kafka cluster.
+            /// - Parameter source: ``NIOAsyncSequenceProducer/Source`` used for yielding new elements.
+            /// - Parameter subscribedTopicsPointer: Pointer to a list of topics + partition pairs.
+            /// - Parameter logger: A logger.
+            case initializing(
+                pollInterval: Duration,
+                client: KafkaClient,
+                source: Producer.Source,
+                subscribedTopicsPointer: UnsafeMutablePointer<rd_kafka_topic_partition_list_t>,
+                logger: Logger
```
FranzBusch, line 374: Let's not store the logger in here. We can just keep it in a separate prop since it is not state.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -289,72 +271,285 @@ public final class KafkaConsumer {
...
+            /// - Parameter pollInterval: Amount of time between two subsequent polls invocations.
+            /// - Parameter client: Client used for handling the connection to the Kafka cluster.
+            /// - Parameter source: ``NIOAsyncSequenceProducer/Source`` used for yielding new elements.
+            /// - Parameter subscribedTopicsPointer: Pointer to a list of topics + partition pairs.
+            /// - Parameter logger: A logger.
+            case initializing(
+                pollInterval: Duration,
+                client: KafkaClient,
+                source: Producer.Source,
+                subscribedTopicsPointer: UnsafeMutablePointer<rd_kafka_topic_partition_list_t>,
```
FranzBusch, line 373: We should create an abstraction for this so we don't hold `UnsafeMutablePointer`s all the time.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -289,72 +271,285 @@ public final class KafkaConsumer {
...
+            /// We are in the process of initializing the ``KafkaConsumer``,
+            /// though ``subscribe()`` / ``assign()`` have not been invoked.
+            ///
+            /// - Parameter pollInterval: Amount of time between two subsequent polls invocations.
+            /// - Parameter client: Client used for handling the connection to the Kafka cluster.
+            /// - Parameter source: ``NIOAsyncSequenceProducer/Source`` used for yielding new elements.
+            /// - Parameter subscribedTopicsPointer: Pointer to a list of topics + partition pairs.
+            /// - Parameter logger: A logger.
+            case initializing(
+                pollInterval: Duration,
```
FranzBusch, line 370: Doesn't need to be in the state machine since it is a fixed configuration.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -289,72 +271,285 @@ public final class KafkaConsumer {
...
+            /// - Parameter source: ``NIOAsyncSequenceProducer/Source`` used for yielding new elements.
+            /// - Parameter logger: A logger.
+            case pollForAndYieldMessage(
+                pollInterval: Duration,
+                client: KafkaClient,
+                source: Producer.Source,
+                logger: Logger
+            )
+            /// Kill the poll loop.
+            case killPollLoop
```
FranzBusch, line 433: Kill is a bit harsh. Can we call this `terminate` instead?

Sources/SwiftKafka/KafkaConsumer.swift
felixschlegel: So `consumerPoll` `throws` when there is an error with the received message. However, I think we shouldn't close the entire consumer because of, e.g. a message with a faulty payload.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -55,32 +85,26 @@ public struct ConsumerMessagesAsyncSequence: AsyncSequence {
...
+
 /// Receive messages from the Kafka cluster.
 public final class KafkaConsumer {
+    typealias Producer = NIOAsyncSequenceProducer<
+        Result<KafkaConsumerMessage, KafkaError>,
+        NoBackPressure,
+        ShutdownOnTerminate
+    >
+    /// Time between two consecutive polls.
+    private var pollInterval: Duration
```
FranzBusch, line 98: Why is this not part of the `KafkaConsumerConfiguration`?

Sources/SwiftKafka/KafkaConsumer.swift
FranzBusch: Hm, I think this is pretty bad. If we get a bad message something seriously went wrong. At least that is what I expect. Can you investigate when this could happen and document it here. IMO it is weird that our sequence contains a `Result<>`, I would like to get rid of that.

Sources/SwiftKafka/RDKafka/RDKafkaTopicPartitionList.swift
```diff
@@ -0,0 +1,60 @@
...
+// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+
+/// Swift wrapper type for `rd_kafka_topic_partition_list_t`.
+class RDKafkaTopicPartitionList {
```
FranzBusch, line 18:
```suggestion
final class RDKafkaTopicPartitionList {
```

Sources/SwiftKafka/KafkaConsumer.swift
felixschlegel: Wanted to do that in a separate PR as this also affects `KafkaProducer`, but yeah I can do it in this one already

FranzBusch: Separate PR is fine

felixschlegel: Too late now 😅

FranzBusch (approved):

Issue #64, opened 2023-06-15
Internalise properties of `KafkaProducer` into its `state`


felixschlegel:
Internalise properties of `KafkaProducer` into its `state` using `enum` associated values as well as possible

> @FranzBusch : https://github.com/swift-server/swift-kafka-gsoc/pull/63#discussion_r1230907351

ssikka100:
This is done. The producer state machine already uses enum associated values — client, messageIDCounter, source, and topicHandles are carried inside .started(...), and client/source in .finishing(...). Properties are only accessible in the correct lifecycle state. Closing.

PR #67, opened 2023-06-22
Refactor `KafkaProducer`

felixschlegel:
> **Fixes**: #64 

### Motiviation:

* align `KafkaProducer` more with proposed changes to `KafkaConsumer`
* `AsyncStream` was not handling `AsyncSequence` termination handling as
  we wanted it to, so revert back to use `NIOAsyncSequenceProducer`

### Modifications:

* make `KafkaProducer` `final class` instead of `actor`
* `KafkaProducer`: use `NIOAsyncSequenceProducer` instead of
  `AsyncSequence` for better termination handling -> shutdown
  `KafkaProducer` on termination of the `AsyncSequence`
* introduce `StateMachine` to `KafkaProducer`
* move the internal state of `KafkaProducer` to `KafkaProducer.StateMachine`
* remove unused `await` expressions when accessing `KafkaProducer`
* update tests
* update `README`

### DIFF
--- a/Sources/SwiftKafka/KafkaProducer.swift
+++ b/Sources/SwiftKafka/KafkaProducer.swift
@@ -14,16 +14,56 @@
 
 import Crdkafka
 import Logging
+import NIOConcurrencyHelpers
+import NIOCore
+
+/// `NIOAsyncSequenceProducerBackPressureStrategy` that always returns true.
+struct NoBackPressure: NIOAsyncSequenceProducerBackPressureStrategy {
+    func didYield(bufferDepth: Int) -> Bool { true }
+    func didConsume(bufferDepth: Int) -> Bool { true }
+}
+
+// MARK: - ShutDownOnTerminate
+
+/// `NIOAsyncSequenceProducerDelegate` that terminates the shuts the producer down when
+/// `didTerminate()` is invoked.
+struct ShutdownOnTerminate: @unchecked Sendable { // We can do that because our stored propery is protected by a lock
+    let stateMachine: NIOLockedValueBox<KafkaProducer.StateMachine>
+}
+
+extension ShutdownOnTerminate: NIOAsyncSequenceProducerDelegate {
+    func produceMore() {
+        // No back pressure
+        return
+    }
+
+    func didTerminate() {
+        let action = self.stateMachine.withLockedValue { $0.finish() }
+        switch action {
+        case .shutdownGracefullyAndFinishSource(let client, let source, let topicHandles):
+            Task {
+                await KafkaProducer._shutDownGracefully(
+                    client: client,
+                    topicHandles: topicHandles,
+                    source: source,
+                    timeout: 10000
+                )
+            }
+        case .none:
+            return
+        }
+    }
+}
 
 /// `AsyncSequence` implementation for handling messages acknowledged by the Kafka cluster (``KafkaAcknowledgedMessage``).
 public struct KafkaMessageAcknowledgements: AsyncSequence {
     public typealias Element = Result<KafkaAcknowledgedMessage, KafkaAcknowledgedMessageError>
-    typealias WrappedSequence = AsyncStream<Element>
+    typealias WrappedSequence = NIOAsyncSequenceProducer<Element, NoBackPressure, ShutdownOnTerminate>
     let wrappedSequence: WrappedSequence
 
     /// `AsynceIteratorProtocol` implementation for handling messages acknowledged by the Kafka cluster (``KafkaAcknowledgedMessage``).
     public struct AcknowledgedMessagesAsyncIterator: AsyncIteratorProtocol {
-        var wrappedIterator: AsyncStream<Element>.AsyncIterator
+        var wrappedIterator: WrappedSequence.AsyncIterator
 
         public mutating func next() async -> Element? {
             await self.wrappedIterator.next()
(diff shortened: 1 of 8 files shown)


Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -14,17 +14,56 @@
...
+    func produceMore() {
+        // No back pressure
+        return
+    }
+
+    func didTerminate() {
+        let action = self.stateMachine.withLockedValue { $0.finish() }
+        switch action {
+        case .shutdownGracefullyAndFinishSource(let client, let source, let topicHandles):
+            Task {
```
FranzBusch, line 44: We can't do this. Can we instead do the same thing for flush as you did with commitSync and close i.e. using the queue and a callback. Then we can use a continuation somewhere. We somehow need to get this termination event over to the `KafkaProducer`s run method probably.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -255,36 +310,230 @@ public actor KafkaProducer {
...
+            ///
+            /// - Parameter messageIDCounter:Used to incrementally assign unique IDs to messages.
+            /// - Parameter client: Client used for handling the connection to the Kafka cluster.
+            /// - Parameter source: ``NIOAsyncSequenceProducer/Source`` used for yielding new elements.
+            /// - Parameter topicHandles: Dictionary containing all topic names with their respective `rd_kafka_topic_t` pointer.
+            case started(
+                client: KafkaClient,
+                messageIDCounter: UInt,
+                source: Producer.Source?,
+                topicHandles: [String: OpaquePointer]
```
FranzBusch, line 372: Could we abstract the handles into a class as well so we avoid the `OpaquePointer` here?

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -255,36 +310,230 @@ public actor KafkaProducer {
...
+                topicHandles: [:]
+            )
+        }
+
+        /// Action to be taken when wanting to poll.
+        enum PollLoopAction {
+            /// Poll client for new consumer messages.
+            case poll(client: KafkaClient)
+            /// Kill the poll loop.
+            case killPollLoop
```
FranzBusch, line 403: Can we use a different word than kill here like terminate?

FranzBusch (changes requested):

Sources/SwiftKafka/KafkaProducer.swift
felixschlegel: Unfortunately, I haven't found any async alternative to [`rd_kafka_flush`](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#aaff06c4372bce917c17f3c1a5d8b205d) like there is with [`rd_kafka_consumer_close_queue`](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#a9dd5c18bdfed81c8847b259f0a8d498d) or [`rd_kafka_commit_queue`](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#af76a6a73baa9c2621536e3f6882a3c1a).

What I have done however is get rid of [`rd_kafka_flush`](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#aaff06c4372bce917c17f3c1a5d8b205d) and imitate this method as good as possible using our `StateMachine` and our run loop. However, what I have done is very basic compared to the [actual `rd_kafka_flush` implementation](https://github.com/confluentinc/librdkafka/blob/d174c0dddc1a6c17a93b9b3e1d10e5cf4b2b61b2/src/rdkafka.c#L4346). I am basically just polling until there are no more events in `librdkafka`'s outgoing queue when `KafkaProducer.shutdownGracefully` has been invoked. Atm there are no `timeouts` whatsoever. Please let me know what you think.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -140,82 +152,72 @@ public actor KafkaProducer {
...
-            }
-        }
-
-        for (_, topicHandle) in self.topicHandles {
-            rd_kafka_topic_destroy(topicHandle)
-        }
-
-        self.state = .shutDown
+    public func shutdownGracefully() {
+        self.stateMachine.withLockedValue { $0.finish() }
```
FranzBusch, line 202: Let's call this `triggerGracefulShutdown` since we are actually not waiting for the shutdown here. Once we adopt lifecycle we should make this method non `public`

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -228,68 +230,161 @@ public actor KafkaProducer {
...
+                    topicHandles: topicHandles
                 )
-                // rd_kafka_topic_new deallocates topic config object
+                return .send(
+                    client: client,
+                    newMessageID: newMessageID,
+                    topicHandles: topicHandles
+                )
+            case .flushing, .finished:
+                throw KafkaError.connectionClosed(reason: "Tried to produce a message with a closed producer")
```
FranzBusch, line 372: Can we create a new separate error here when we flushing which indicates that we are currently shutting down and flushing please

FranzBusch (approved):

PR #71, opened 2023-06-23
Dedicated `KafkaProducerMessageID` type

felixschlegel:
> **Warning**: this can potentially be a breaking change as message ids
> will lose their chronological order in the public API after this
> change.

### Motivation:

We don't want to expose `UInt` for `KafkaProducer`'s message ids.

### Modifications:

* create a new `struct` `KafkaProducerMessageID` that is internally
  backed by `UInt` (can also be `UUID` or similar in the future)
* update tests

### DIFF
--- a/Sources/SwiftKafka/KafkaProducerMessageID.swift
+++ b/Sources/SwiftKafka/KafkaProducerMessageID.swift
@@ -34,3 +34,11 @@ extension KafkaProducerMessageID: CustomStringConvertible {
 // MARK: - KafkaProducerMessageID + Hashable
 
 extension KafkaProducerMessageID: Hashable {}
+
+// MARK: - KafkaProducerMessageID + Comparable
+
+extension KafkaProducerMessageID: Comparable {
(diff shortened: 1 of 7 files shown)


Sources/SwiftKafka/KafkaProducerMessageID.swift
```diff
@@ -0,0 +1,36 @@
...
+
+extension KafkaProducerMessageID: CustomStringConvertible {
+    public var description: String {
+        return String(self.rawValue)
+    }
+}
+
+// MARK: - KafkaProducerMessageID + Hashable
+
+extension KafkaProducerMessageID: Hashable {}
```
FranzBusch, line 36: We should probably add a `Comparable` here as well

felixschlegel: Yes, certainly, have implemented this now. This way our users can still reconstruct the chronological order in which they enqueued messages for sending 😄

FranzBusch (approved):

PR #72, opened 2023-06-26
`KafkaConsumer`: move `ConsumptionStrategy` to `init`

felixschlegel:
### Motivation:

One of our internal adopters asked for this as they were using mutliple
`KafkaConsumer`s that all shared the same `KafkaConsumerConfiguration`
but had different `ConsumptionStrategy`s. This meant they had to
duplicate their `KafkaConsumerConfiguration` just to accomodate for
different consumption strategies.

### Modifications:

* move `ConsumptionStrategy` for `KafkaConsumer` from
  `KafkaConsumerConfiguration` to `KafkaConsumer.init`
* update `README`
* update tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,15 +48,15 @@ After initializing the `KafkaConsumer` with a topic-partition pair to read from,
 
 ```swift
 let config = KafkaConsumerConfiguration(
-    consumptionStrategy: .partition(
-        topic: "topic-name",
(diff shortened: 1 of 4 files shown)


FranzBusch (commented):
I am unsure about this change. Having all configurations in a single struct should be fine. If somebody wants to have a single thing different they can just mutate that one property. Since the configuration is a struct it will just CoW it.

PR #68, opened 2023-06-22
Fix: `KafkaConsumer.commitSync` should not block

felixschlegel:
> **Info**: This PR sits on top of #66

### Motivation

Currently our invocation to `rd_kafka_commit` inside of
`KafkaCosumer.commitSync` is blocking a cooperative thread.
This PR aims to make `KafkaCosumer.commitSync` non-blocking by using the
callback-based commit API.

### Modifications:

* move `commitSync` logic to `KafkaClient`
* replace the blocking invocation to [rd_kafka_commit](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#ab96539928328f14c3c9177ea0c896c87) with a callback-based invocation to [rd_kafka_commit_queue](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#af76a6a73baa9c2621536e3f6882a3c1a)
which is then wrapped inside a `withAsyncThrowingContinuation` statement

PR #73, opened 2023-06-27
Fix: `KafkaClient.closeConsumer` should not block

felixschlegel:
> **Info**: this PR sits of top of #68

### Motivation:

[rd_kakfa_consumer_close](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#a37b54d329e12d745889defe96e7d043d)
was blocking. This PR proposes using the
[rd_kakfa_consumer_close_queue](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#a9dd5c18bdfed81c8847b259f0a8d498d)
API which is non-blocking and served through the normal poll loop.

### Modifications:

* `KafkaClient.consumerClose`: use `rd_kakfa_consumer_close_queue` in
   favour of `rd_kakfa_consumer_close`
* create a new variable `KafkaClient.isConsumerClosed` that indicates
  if the poll loop needs to continue polling or if it can stop running
* updated state management in `KafkaConsumer` to accomodate for polling
  when the `KafkaConsumer` is in the process of closing

### Result:

Calling `KafkaClient.consumerClose` is not blocking anymore

### DIFF
--- a/Sources/SwiftKafka/KafkaConsumer.swift
+++ b/Sources/SwiftKafka/KafkaConsumer.swift
@@ -40,10 +40,10 @@ extension ShutdownOnTerminate: NIOAsyncSequenceProducerDelegate {
     }
 
     func didTerminate() {
-        // Duplicate of _shutdownGracefully
+        // Duplicate of _triggerGracefulShutdown
         let action = self.stateMachine.withLockedValue { $0.finish() }
         switch action {
-        case .shutdownGracefullyAndFinishSource(let client, let source):
+        case .triggerGracefulShutdownAndFinishSource(let client, let source):
             source.finish()
 
             do {
@@ -151,7 +151,7 @@ public final class KafkaConsumer {
     }
 
     deinit {
-        self.shutdownGracefully()
+        self.triggerGracefulShutdown()
     }
 
     /// Subscribe to the given list of `topics`.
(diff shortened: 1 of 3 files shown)


Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -243,29 +236,7 @@ public final class KafkaConsumer {
...
-                        listPointer,
-                        0
-                    ) // Blocks until commiting the offset is done
-                    // -> Will be resolved by: https://github.com/swift-server/swift-kafka-gsoc/pull/68
-                }
-            }
-            guard result == RD_KAFKA_RESP_ERR_NO_ERROR else {
-                throw KafkaError.rdKafkaError(wrapping: result)
-            }
+            try await client.commitSync(message)
```
FranzBusch, line 239: The underlying APIs don't support cancellation. Let's call this out in the docs of this method.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -337,7 +308,9 @@ extension KafkaConsumer {
                 source: Producer.Source
             )
             /// The ``KafkaConsumer`` has been closed.
-            case finished
+            ///
+            /// - Parameter client: Client used for handling the connection to the Kafka cluster.
+            case finished(client: KafkaClient)
```
FranzBusch, line 313: That is a bit weird. We should make sure that the end state actually doesn't retain the client so that we can deinit it. Maybe we need a new state here.

FranzBusch (changes requested):

FranzBusch (approved):

PR #82, opened 2023-07-06
`KafkaProducer`: replace callbacks with `eventPoll`

felixschlegel:
### Motivation:

Currently we serve `librdkafka` logs and delivery reports through
callbacks. In the future we also want to get notified about `librdkafka`
triggered errors, statistics and rebalances. As having the callbacks
adds a lot of overhead e.g. by having `Unmanaged` references and having
to wrap Swift closures in C callbacks, we rather want to have an
`eventPoll` method that regularly polls for all sorts of events and
handles them inline.

### Modifications:

* `KafkaClient`
    * remove `poll` method
    * add new method `eventPoll`
    * move mapping from `librdkafka` logs to `swift-log` to
      `KafkaClient.eventPoll`
* `KafkaProducer`:
    * receive acknowledgements through `KafkaClient.eventPoll` instead of using
      `KafkaClient.poll` to trigger event callbacks
    * replace `deliveryReport` callback and handle deliveryReport event directly in
      `KafkaProducer`
* `RDKafkaConfig`
    * remove `logging` logic
* `KafkaProducerTests`:
    * add `testProducerLog` that asserts that `librdkafka` logs get
      forwarded to `swift-log`
* move `convertMessageToAcknowledgementResult` from `RDKafkaConfig` to
  `KafkaClient`

### DIFF
--- a/Sources/SwiftKafka/KafkaClient.swift
+++ b/Sources/SwiftKafka/KafkaClient.swift
@@ -87,63 +87,100 @@ final class KafkaClient: Sendable {
     }
 
     /// Swift wrapper for events from `librdkafka`'s event queue.
-    enum RDKafkaEvent {
+    enum KafkaEvent {
         case deliveryReport(results: [Result<KafkaAcknowledgedMessage, KafkaAcknowledgedMessageError>])
     }
 
     /// Poll the event `rd_kafka_queue_t` for new events.
     ///
     /// - Parameter maxEvents:Maximum number of events to serve in one invocation.
-    func eventPoll(maxEvents: Int = 1000) -> [RDKafkaEvent] {
-        var events = [RDKafkaEvent]()
+    func eventPoll(maxEvents: Int = 100) -> [KafkaEvent] {
+        var events = [KafkaEvent]()
+        events.reserveCapacity(maxEvents)
 
-        getEventsLoop: for _ in 0..<maxEvents {
+        for _ in 0..<maxEvents {
             let event = rd_kafka_queue_poll(self.mainQueue, 0)
-            let eventType = rd_kafka_event_type(event)
+            defer { rd_kafka_event_destroy(event) }
+
+            let rdEventType = rd_kafka_event_type(event)
+            guard let eventType = RDKafkaEvent(rawValue: rdEventType) else {
+                fatalError("Unsupported event type: \(rdEventType)")
+            }
 
             switch eventType {
-            case RD_KAFKA_EVENT_DR:
-                var deliveryReportResults = [Result<KafkaAcknowledgedMessage, KafkaAcknowledgedMessageError>]()
-                while let messagePointer = rd_kafka_event_message_next(event) {
-                    guard let message = Self.convertMessageToAcknowledgementResult(messagePointer: messagePointer) else {
-                        continue
-                    }
-                    deliveryReportResults.append(message)
-                }
-                events.append(.deliveryReport(results: deliveryReportResults))
-            case RD_KAFKA_EVENT_LOG:
-                var faculty: UnsafePointer<CChar>?
-                var buffer: UnsafePointer<CChar>?
-                var level: Int32 = 0
-                if rd_kafka_event_log(event, &faculty, &buffer, &level) == 0 {
-                    if let faculty, let buffer {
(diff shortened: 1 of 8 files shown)


Sources/SwiftKafka/KafkaClient.swift
```diff
@@ -83,18 +86,64 @@ final class KafkaClient: Sendable {
...
-    /// (in milliseconds) that the call will block waiting for events.
-    /// For non-blocking calls, provide 0 as `timeout`.
-    /// To wait indefinitely for an event, provide -1.
-    /// - Returns: The number of events served.
-    @discardableResult
-    func poll(timeout: Int32) -> Int32 {
-        return rd_kafka_poll(self.kafkaHandle, timeout)
+    /// - Parameter maxEvents:Maximum number of events to serve in one invocation.
+    func eventPoll(maxEvents: Int = 1000) -> [RDKafkaEvent] {
+        var events = [RDKafkaEvent]()
```
FranzBusch, line 98: Can we decrease the maxEvents number and call `reserveCapacity` on the events array so we don't realloc the whole time

Sources/SwiftKafka/KafkaClient.swift
```diff
@@ -83,18 +86,64 @@ final class KafkaClient: Sendable {
...
+
+        getEventsLoop: for _ in 0..<maxEvents {
```
FranzBusch, line 100: I am not a big fan of named loops. I don't think we need this here since we can just return in the case where we call `break getEventsLoop` right now.

Sources/SwiftKafka/KafkaClient.swift
```diff
@@ -83,18 +86,64 @@ final class KafkaClient: Sendable {
...
+            let event = rd_kafka_queue_poll(self.mainQueue, 0)
```
FranzBusch, line 101: Can we add a `defer` here that destroys the event?

Sources/SwiftKafka/KafkaClient.swift
```diff
@@ -83,18 +86,64 @@ final class KafkaClient: Sendable {
...
+            let eventType = rd_kafka_event_type(event)
+
+            switch eventType {
+            case RD_KAFKA_EVENT_DR:
```
FranzBusch, line 105: Can we split out each case into a separate private method

Sources/SwiftKafka/KafkaClient.swift
```diff
@@ -83,18 +86,64 @@ final class KafkaClient: Sendable {
...
+                var deliveryReportResults = [Result<KafkaAcknowledgedMessage, KafkaAcknowledgedMessageError>]()
```
FranzBusch, line 106: Can we reserveCapacity here for the count of messages

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -149,15 +147,7 @@ public final class KafkaProducer: Service, Sendable {
...
-            deliveryReportCallback: { [logger, source] messageResult in
-                guard let messageResult else {
-                    logger.error("Could not resolve acknowledged message")
-                    return
-                }
-
-                // Ignore YieldResult as we don't support back pressure in KafkaProducer
-                _ = source.yield(messageResult)
-            },
+            events: RD_KAFKA_EVENT_LOG | RD_KAFKA_EVENT_DR,
```
FranzBusch, line 150: Can we abstract this into a Swift enum so we don't have to deal with C symbols here

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -193,8 +183,15 @@ public final class KafkaProducer: Service, Sendable {
         while !Task.isCancelled {
             let nextAction = self.stateMachine.withLockedValue { $0.nextPollLoopAction() }
             switch nextAction {
-            case .poll(let client):
-                client.poll(timeout: 0)
+            case .poll(let client, let source):
+                let events = client.eventPoll()
```
FranzBusch, line 187: Maybe I am missing something but in the case where our async sequence is shutdown we still have to poll right? Otherwise kafka will just queue up the events.

felixschlegel: Yes, this is implemented in another PR (#84) though 😄

PR #84, opened 2023-07-07
`KafkaProducer`: Drop acks on sequence termination

felixschlegel:
### Motivation:

We do not to buffer any incoming messages when the asynchronous
sequences we expose are terminated.

This PR proposes adding a new state to `KafkaProducer` that drops any incoming acknowledgements after the asynchronous
sequence has been terminated.

> **Note**: we are not shutting down entirely as terminating the `run()`
loop early is seen as an error is `swift-service-lifecycle`.

### Modifications:

* add new `NIOAsyncSequenceProducerDelegate` `KafkaProducerCloseOnTerminate`
* `KafkaProducer`: add new stateMachine logic
  `stopConsuming` that does not stop the poll loop but drops all
  incoming messages / acknowledgements
* add test
  `KafkaProducerTests.testSendFailsAfterTerminatingAcknowledgementSequence`

### DIFF
--- a/Sources/SwiftKafka/KafkaConsumer.swift
+++ b/Sources/SwiftKafka/KafkaConsumer.swift
@@ -18,6 +18,14 @@ import NIOConcurrencyHelpers
 import NIOCore
 import ServiceLifecycle
 
+// MARK: - NoDelegate
+
+// `NIOAsyncSequenceProducerDelegate` that does nothing.
+internal struct NoDelegate: NIOAsyncSequenceProducerDelegate {
+    func produceMore() {}
+    func didTerminate() {}
+}
+
 // MARK: - KafkaConsumerMessages
 
 /// `AsyncSequence` implementation for handling messages received from the Kafka cluster (``KafkaConsumerMessage``).
--- a/Sources/SwiftKafka/KafkaProducer.swift
+++ b/Sources/SwiftKafka/KafkaProducer.swift
@@ -22,7 +22,7 @@ import ServiceLifecycle
 
 /// `NIOAsyncSequenceProducerDelegate` that terminates the closes the producer when
 /// `didTerminate()` is invoked.
-internal struct KafkaProducerCloseOnTerminate: @unchecked Sendable { // We can do that because our stored propery is protected by a lock
+internal struct KafkaProducerCloseOnTerminate: Sendable {
     let stateMachine: NIOLockedValueBox<KafkaProducer.StateMachine>
 }
 
@@ -32,7 +32,13 @@ extension KafkaProducerCloseOnTerminate: NIOAsyncSequenceProducerDelegate {
     }
 
     func didTerminate() {
-        self.stateMachine.withLockedValue { $0.stopConsuming() }
+        let action = self.stateMachine.withLockedValue { $0.stopConsuming() }
+        switch action {
+        case .finishSource(let source):
+            source?.finish()
+        case .none:
+            break
+        }
     }
 }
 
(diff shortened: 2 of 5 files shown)


Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -360,13 +392,53 @@ extension KafkaProducer {
...
+        mutating func stopConsuming() {
+            switch self.state {
+            case .uninitialized:
+                fatalError("\(#function) invoked while still in state \(self.state)")
+            case .started(let client, _, _, _):
+                self.state = .consumptionStopped(client: client)
+            case .flushing(let client, _):
+                // Setting source to nil to prevent incoming acknowledgements from buffering in `source`
+                self.state = .flushing(client: client, source: nil)
+            case .consumptionStopped, .finished:
```
FranzBusch, line 415: Can we fatalError in the `consumptionStopped` case as well please.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -268,6 +289,11 @@ extension KafkaConsumer {
                 client: KafkaClient,
                 source: Producer.Source
             )
+            /// Consumer is still running but the messages asynchronous sequence was terminated.
+            /// All incoming message will be dropped.
+            ///
+            /// - Parameter client: Client used for handling the connection to the Kafka cluster.
+            case consumptionStopped(client: KafkaClient)
```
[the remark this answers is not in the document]
felixschlegel: Oh yeah stupid mistake, I was too caught up in the drop incoming messages narrative and did not consider that dropping consumer messages might not be the brightest idea 😅

felixschlegel:

felixschlegel:
Hey @FranzBusch,

I have changed this PR to only affect the `KafkaProducer` for now.

> I don't think that state makes sense for the consumer. This can lead to the user dropping the iterator by accident and we will still consume messages and drop them on the floor which they never got. In the end, this would screw with their offset because for Kafka it looks like the consumed it.

This question made me discover that `librdkafka` automatically stores commit offsets every time we invoke `rd_kafka_consumer_poll` (or `rd_kafka_queue_poll` for the new planned event loop).

My plan now is to create a "proxy" for `enable.auto.commit` and handle committing ourselves so we gain full control of the message offsets we commit.

Tests/SwiftKafkaTests/KafkaProducerTests.swift
```diff
@@ -210,6 +210,56 @@ final class KafkaProducerTests: XCTestCase {
...
+        )
+
+        try await withThrowingTaskGroup(of: Void.self) { group in
+            // Run Task
+            group.addTask {
+                try await serviceGroup.run()
+            }
+
+            // Test Task
+            group.addTask {
```
FranzBusch, line 229: You don't need a separate task for this.

Tests/SwiftKafkaTests/KafkaProducerTests.swift
```diff
@@ -210,6 +210,56 @@ final class KafkaProducerTests: XCTestCase {
...
+                    key: "key2",
+                    value: "Hello, San Diego!"
+                )
+
+                try producer.send(message1)
+
+                // Terminate the acknowledgements sequence by deallocating its AsyncIterator
+                var iter: KafkaMessageAcknowledgements.AsyncIterator? = acks.makeAsyncIterator()
+                _ = iter
+                iter = nil
```
FranzBusch, line 246: Please spell `iterator` out here

Tests/SwiftKafkaTests/KafkaProducerTests.swift
```diff
@@ -210,6 +210,56 @@ final class KafkaProducerTests: XCTestCase {
         }
     }
 
+    func testSendFailsAfterTerminatingAcknowledgementSequence() async throws {
+        let (producer, acks) = try KafkaProducer.makeProducerWithAcknowledgements(config: self.config, logger: .kafkaTest)
```
FranzBusch, line 214: Please spell acks out here

FranzBusch (approved):

PR #85, opened 2023-07-10
`KafkaConsumer`: drop messages on sequence termination

felixschlegel:
### Motivation:

We want to avoid running out of memory when the user stops reading the
`KafkaConsumer.messages` `AsyncSequence`. Therefore we drop any incoming
messages that we receive after the `AsyncSequence` was terminated.

### Modifications:

* `KafkaConsumer`
    * enter idle polling state when `KafkaConsumer.messages`
      `AsyncSequence` was terminated to ensure that the `func run()`
      async does not return early and result in an error for its `ServiceGroup`
* `KafkaConsumerConfiguration`:
    * proxy `enable.auto.commit` and make it `false` by default
      Reason: our library should be in charge of deciding when a message
      offset is to be commited because only we know if our library user received the message through the
      `KafkaConsumer.messages` `AsyncSequence`
* `KafkaSharedConfiguration.partition`: document default value of
  parameter `offset`
* add new test `SwiftKafkaTests.testCommittedOffsetsAreCorrect()`

### DIFF
--- a/Sources/SwiftKafka/KafkaConsumer.swift
+++ b/Sources/SwiftKafka/KafkaConsumer.swift
@@ -22,7 +22,7 @@ import ServiceLifecycle
 
 /// `NIOAsyncSequenceProducerDelegate` that terminates the closes the producer when
 /// `didTerminate()` is invoked.
-internal struct KafkaConsumerCloseOnTerminate: @unchecked Sendable { // We can do that because our stored propery is protected by a lock
+internal struct KafkaConsumerCloseOnTerminate: Sendable {
     let stateMachine: NIOLockedValueBox<KafkaConsumer.StateMachine>
 }
 
--- a/Sources/SwiftKafka/KafkaConsumer.swift
+++ b/Sources/SwiftKafka/KafkaConsumer.swift
@@ -237,7 +237,7 @@ public final class KafkaConsumer: Sendable, Service {
                 // Ignore poll result.
                 // We are just polling to serve any remaining events queued inside of `librdkafka`.
                 // All remaining queued consumer messages will get dropped and not be committed (marked as read).
-                _ = try client.eventPoll()
+                _ = client.eventPoll()
                 try await Task.sleep(for: self.config.pollInterval)
             case .terminatePollLoop:
                 return
--- a/Sources/SwiftKafka/KafkaConsumer.swift
+++ b/Sources/SwiftKafka/KafkaConsumer.swift
@@ -18,12 +18,22 @@ import NIOConcurrencyHelpers
 import NIOCore
 import ServiceLifecycle
 
-// MARK: - NoDelegate
+// MARK: - KafkaConsumerCloseOnTerminate
 
-// `NIOAsyncSequenceProducerDelegate` that does nothing.
-internal struct NoDelegate: NIOAsyncSequenceProducerDelegate {
-    func produceMore() {}
-    func didTerminate() {}
+/// `NIOAsyncSequenceProducerDelegate` that terminates the closes the producer when
+/// `didTerminate()` is invoked.
+internal struct KafkaConsumerCloseOnTerminate: @unchecked Sendable { // We can do that because our stored propery is protected by a lock
+    let stateMachine: NIOLockedValueBox<KafkaConsumer.StateMachine>
+}
+
+extension KafkaConsumerCloseOnTerminate: NIOAsyncSequenceProducerDelegate {
+    func produceMore() {
+        return // No back pressure
+    }
+
+    func didTerminate() {
+        self.stateMachine.withLockedValue { $0.stopConsuming() }
+    }
 }
 
 // MARK: - KafkaConsumerMessages
@@ -32,7 +42,7 @@ internal struct NoDelegate: NIOAsyncSequenceProducerDelegate {
 public struct KafkaConsumerMessages: Sendable, AsyncSequence {
     public typealias Element = KafkaConsumerMessage
     typealias BackPressureStrategy = NIOAsyncSequenceProducerBackPressureStrategies.NoBackPressure
-    typealias WrappedSequence = NIOAsyncSequenceProducer<Element, BackPressureStrategy, NoDelegate>
+    typealias WrappedSequence = NIOAsyncSequenceProducer<Element, BackPressureStrategy, KafkaConsumerCloseOnTerminate>
     let wrappedSequence: WrappedSequence
 
     /// `AsynceIteratorProtocol` implementation for handling messages received from the Kafka cluster (``KafkaConsumerMessage``).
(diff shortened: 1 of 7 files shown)


FranzBusch (changes requested):
Let's do this slightly different since we shouldn't commit on every read message. Rather let's update the offset on every `next()` call and disable the auto store offset configuration.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -18,34 +18,71 @@ import NIOConcurrencyHelpers
...
+            switch action {
+            case .storeOffset(let client):
+                do {
+                    try client.storeMessageOffset(element)
+                } catch {
+                    self.logger.error("""
+                    Failed committing offset '\(element.offset)' for partition '\(element.partition)'\
+                    of topic '\(element.topic)'
+                    """)
+                    return nil // Finishes sequence
```
FranzBusch, line 73: IMO we should throw the error here instead of silently failing. Having this sequence be a throwing one is totally fine.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -183,7 +223,7 @@ public final class KafkaConsumer: Sendable, Service {
                     throw error
                 }
                 try await Task.sleep(for: self.config.pollInterval)
-            case .pollUntilClosed(let client):
+            case .pollWithoutYield(let client):
```
FranzBusch, line 226: This also means that we are polling the queue and all messages will just get dropped and never committed. I think this is the sane thing to do. We should add this as a comment here.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -281,6 +321,15 @@ extension KafkaConsumer {
...
                 source: Producer.Source
             )
+            /// Consumer is still running but the messages asynchronous sequence was terminated.
+            /// All incoming messages will be dropped.
+            ///
+            /// - Parameter client: Client used for handling the connection to the Kafka cluster.
+            /// - Parameter source: ``NIOAsyncSequenceProducer/Source`` used for yielding new elements.
+            case consumptionStopped(
+                client: KafkaClient,
+                source: Producer.Source
```
FranzBusch, line 331: Why do we still need the `source` here?

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -371,11 +423,51 @@ extension KafkaConsumer {
...
                 return .setUpConnection(client: client)
-            case .consuming, .finishing, .finished:
+            case .consuming, .consumptionStopped, .finishing, .finished:
                 fatalError("\(#function) should only be invoked upon initialization of KafkaConsumer")
             }
         }
 
+        /// The messages asynchronous sequence was terminated.
+        /// All incoming messages will be dropped.
+        mutating func stopConsuming() {
```
FranzBusch, line 433: Can we call this `messageSequenceTerminated`

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -371,11 +423,51 @@ extension KafkaConsumer {
...
+
+        /// Action to take when wanting to store a message offset (to be auto-committed by `librdkafka`).
+        enum StoreOffsetAction {
+            /// Store the message offset with the given `client`.
+            /// - Parameter client: Client used for handling the connection to the Kafka cluster.
+            case storeOffset(client: KafkaClient)
+        }
+
+        /// Get action to take when wanting to store a message offset (to be auto-committed by `librdkafka`).
+        func storeOffset() -> StoreOffsetAction {
```
FranzBusch, line 456: Can we call this `messageSequenceNext` and also rename the action. It is better to name those methods how the places are called and not with the intended outcome.

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -18,34 +18,71 @@ import NIOConcurrencyHelpers
...
+
+            let action = self.stateMachine.withLockedValue { $0.storeOffset() }
+            switch action {
+            case .storeOffset(let client):
+                do {
+                    try client.storeMessageOffset(element)
+                } catch {
+                    self.logger.error("""
+                    Failed committing offset '\(element.offset)' for partition '\(element.partition)'\
+                    of topic '\(element.topic)'
```
glbrntt, line 71: You need to add a space before the `\`:

```suggestion
                    Failed committing offset '\(element.offset)' for partition '\(element.partition)' \
                    of topic '\(element.topic)'
```

Sources/SwiftKafka/KafkaConsumer.swift
glbrntt, line 73: The wrapped iterator doesn't know that we've ended which presumably means we _could_ call `next()` again and actually get a value. This doesn't seem desirable!

Sources/SwiftKafka/KafkaConsumer.swift
glbrntt: I disagree on this one 😅 -- why do you think it is better?

I think it's better if the action you perform on the state machine is named after the thing you want to do as it makes the call site easier to understand. If `messageSequenceNext()` calls `state.messageSequenceNext()` I have no idea what's actually happening. If it calls `state.storeOffet()` I have _some_ idea of what's happening.

A point against naming it against the call site is that it's also just unnecessary coupling.

Sources/SwiftKafka/KafkaConsumer.swift
felixschlegel: We don't 😄

Sources/SwiftKafka/KafkaConsumer.swift
felixschlegel: The `AsyncSequence` `throws` now 👍

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -18,45 +18,76 @@ import NIOConcurrencyHelpers
...
+            guard let element = try await self.wrappedIterator.next() else {
+                return nil
+            }
+
+            let action = self.stateMachine.withLockedValue { $0.storeOffset() }
+            switch action {
+            case .storeOffset(let client):
+                try client.storeMessageOffset(element)
+            }
+            return element
```
FranzBusch, line 70: We still have the issue that we could potentially return something after throwing here. We need to nil out the `wrappedIterator` after throwing or returning nil to make sure we deallocate it.

Sources/SwiftKafka/KafkaConsumer.swift
FranzBusch: I am a bit torn on this one. I do think naming it like the call site is beneficial because in most of these state machines the action we return depends on what we do where. So it is already very coupled.

glbrntt (approved):

FranzBusch (approved):

PR #87, opened 2023-07-12
Refactor `*Configuration` types

felixschlegel:
### Modifications:

* `*Configuration`:
    * remove initializer
    * replace computed properties with normal `var`s
    * make `dictionary` computed property that uses `self` to create a
      `librdkafka` compatible String-based configuration dictionary
* add missing `Sendable` conformance to some types
* rename `KafkaSharedConfiguration` -> `KafkaConfiguration`
* replace `Ms` -> `Milliseconds`
* replace `UInt` -> `Int` in `*Configuration` types


Sources/SwiftKafka/Configuration/KafkaConfiguration.swift
```diff
@@ -0,0 +1,198 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the swift-kafka-gsoc open source project
+//
+// Copyright (c) 2022 Apple Inc. and the swift-kafka-gsoc project authors
```
glbrntt, line 5: 2023

Sources/SwiftKafka/Configuration/KafkaConfiguration.swift
```diff
@@ -0,0 +1,198 @@
...
+
+/// Collection of types used in the configuration structs this library provides.
+public enum KafkaConfiguration {
+    /// The URL of a Kafka broker.
+    public struct Broker: Sendable, Hashable, CustomStringConvertible {
+        /// The host component of the broker URL.
+        public let host: String
+
+        /// The port component of the broker URL.
+        public let port: Int
```
glbrntt, line 23: These should be `vars`

Sources/SwiftKafka/Configuration/KafkaConfiguration.swift
```diff
@@ -0,0 +1,198 @@
...
+// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Collection of types used in the configuration structs this library provides.
+public enum KafkaConfiguration {
+    /// The URL of a Kafka broker.
+    public struct Broker: Sendable, Hashable, CustomStringConvertible {
```
glbrntt, line 18: We need a public `init` too

glbrntt: (and in a bunch of other places!)

Sources/SwiftKafka/Configuration/KafkaConfiguration.swift
```diff
@@ -0,0 +1,198 @@
...
+        /// The port component of the broker URL.
+        public let port: Int
+
+        public var description: String {
+            "\(self.host):\(self.port)"
+        }
+    }
+
+    /// Message options.
+    public struct MessageOptions: Sendable, Hashable {
```
glbrntt, line 31: Can you document the defaults too?

Sources/SwiftKafka/Configuration/KafkaConfiguration.swift
```diff
@@ -0,0 +1,198 @@
...
+        /// Path to client's private key (PEM) used for authentication.
+        public var keyLocation: String = ""
+
+        /// Private key passphrase (for use with ssl.key.location).
+        public var keyPassword: String = ""
+
+        /// Path to client's public key (PEM) used for authentication.
+        public var certificateLocation: String = ""
+
+        /// File or directory path to CA certificate(s) for verifying the broker's key. Defaults: On Windows the system's CA certificates are automatically looked up in the Windows Root certificate store. On Mac OSX this configuration defaults to probe. It is recommended to install openssl using Homebrew, to provide CA certificates. On Linux install the distribution's ca-certificates package. If OpenSSL is statically linked or ssl.ca.location is set to probe a list of standard paths will be probed and the first one found will be used as the default CA certificate location path. If OpenSSL is dynamically linked the OpenSSL library's default path will be used (see OPENSSLDIR in openssl version -a).
```
glbrntt, line 107: Because I'm a pedant:

```suggestion
        /// File or directory path to CA certificate(s) for verifying the broker's key. Defaults: On Windows the system's CA certificates are automatically looked up in the Windows Root certificate store. On macOS this configuration defaults to probe. It is recommended to install openssl using Homebrew, to provide CA certificates. On Linux install the distribution's ca-certificates package. If OpenSSL is statically linked or ssl.ca.location is set to probe a list of standard paths will be probed and the first one found will be used as the default CA certificate location path. If OpenSSL is dynamically linked the OpenSSL library's default path will be used (see OPENSSLDIR in openssl version -a).
```

Sources/SwiftKafka/Configuration/KafkaConfiguration.swift
```diff
@@ -0,0 +1,198 @@
...
+        public var CALocation: String = ""
+
+        /// Path to CRL for verifying broker's certificate validity.
+        public var CRLLocation: String = ""
+
+        /// Path to client's keystore (PKCS#12) used for authentication.
+        public var keystoreLocation: String = ""
+
+        /// Client's keystore (PKCS#12) password.
+        public var keystorePassword: String = ""
```
glbrntt, line 117: Should these be optional strings rather than empty strings? (Elsewhere they are optional)

glbrntt (changes requested):
I think you're missing a bunch of `public init`s but this broadly looks good otherwise.

Sources/SwiftKafka/Configuration/KafkaConfiguration.swift
felixschlegel: I have reimplemented the way we handle security configurations (#88) making this configuration object obsolete 😄

Sources/SwiftKafka/Configuration/KafkaConfiguration.swift
```diff
@@ -0,0 +1,274 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Collection of types used in the configuration structs this library provides.
+public enum KafkaConfiguration {
+    /// The URL of a Kafka broker.
```
FranzBusch, line 17: NIT: Not a URL. Same for the other comments.

Sources/SwiftKafka/Configuration/KafkaConfiguration.swift
```diff
@@ -0,0 +1,274 @@
...
+        public var keyPassword: String = ""
+
+        /// Path to client's public key (PEM) used for authentication.
+        public var certificateLocation: String = ""
+
+        /// File or directory path to CA certificate(s) for verifying the broker's key. Defaults: On Windows the system's CA certificates are automatically looked up in the Windows Root certificate store. On macOS this configuration defaults to probe. It is recommended to install openssl using Homebrew, to provide CA certificates. On Linux install the distribution's ca-certificates package. If OpenSSL is statically linked or ssl.ca.location is set to probe a list of standard paths will be probed and the first one found will be used as the default CA certificate location path. If OpenSSL is dynamically linked the OpenSSL library's default path will be used (see OPENSSLDIR in openssl version -a).
+        public var CALocation: String = ""
+
+        /// Path to CRL for verifying broker's certificate validity.
+        public var CRLLocation: String = ""
```
FranzBusch, line 187: Can we remove the capital start here? e.g. make ti `caLocation`

Sources/SwiftKafka/Configuration/KafkaConfiguration.swift
```diff
@@ -0,0 +1,274 @@
...
+        public var keystoreLocation: String = ""
+
+        /// Client's keystore (PKCS#12) password.
+        public var keystorePassword: String = ""
+    }
+
+    /// SASL options.
+    public struct SASLOptions: Sendable, Hashable {
+        /// SASL mechanism to use for authentication.
+        public var mechanism: KafkaConfiguration.SASLMechanism?
```
FranzBusch, line 199: We should nest this type into this one right?

felixschlegel: Typed security config implemented in #88

Sources/SwiftKafka/Configuration/KafkaConfiguration.swift
```diff
@@ -0,0 +1,274 @@
...
+
+        /// SASL username for use with the PLAIN and SASL-SCRAM-.. mechanisms.
+        public var username: String?
+
+        /// SASL password for use with the PLAIN and SASL-SCRAM-.. mechanisms.
+        public var password: String?
```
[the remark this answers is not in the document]
felixschlegel: Typed security config implemented in https://github.com/swift-server/swift-kafka-gsoc/pull/88

FranzBusch (approved):

Issue #46, opened 2023-01-22
New Config API: Re-evaluate public API & configuration options


felixschlegel:
* Currently, we expose more or less all `librdkafka` configurations — do we want to narrow it down to the most basic configurations and expand this on user request / provide an unsafe API that allows pro-users to set arbitrary key-value options?
* `librdkafka` exposes configuration options regarding OAuth and SSL keys etc. — does our package even intend to expose that as well?

felixschlegel:
Here is a list of the options we are currently not supporting but might want to support in the future:

(See [librdkafka/CONFIGURATION.md](https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md) for reference)

```
topic.metadata.refresh.fast.cnt
connections.max.idle.ms
api.version.request
api.version.request.timeout.ms
ssl.cipher.suites
ssl.curves.list
ssl.sigalgs.list
ssl.key.pem
ssl_key
ssl.certificate.pem
ssl_certificate
ssl.ca.pem
ssl_ca
ssl.ca.certificate.stores
ssl.providers
ssl.engine.location
ssl.engine.id
ssl_engine_callback_data
enable.ssl.certificate.verification
ssl.endpoint.identification.algorithm
ssl.certificate.verify_cb
sasl.mechanisms
sasl.kerberos.service.name
sasl.kerberos.principal
sasl.kerberos.kinit.cmd
sasl.kerberos.keytab
sasl.kerberos.min.time.before.relogin
sasl.oauthbearer.config
sasl.oauthbearer.method
sasl.oauthbearer.client.id
sasl.oauthbearer.client.secret
sasl.oauthbearer.scope
sasl.oauthbearer.extensions
sasl.oauthbearer.token.endpoint.url
group.instance.id
partition.assignment.strategy
coordinator.query.interval.ms
queued.min.messages
queued.max.messages.kbytes
fetch.message.max.bytes
max.partition.fetch.bytes
fetch.min.bytes
fetch.max.bytes
fetch.error.backoff.ms
check.crcs
client.rack
retry.backoff.ms
compression.codec
compression.type
batch.num.messages
batch.size
Not found: sticky.partitioning.linger.ms
```

felixschlegel:
+ `partition.assignment.strategy`

ssikka100:
[...]

PR #88, opened 2023-07-14
Type-safe Security Configuration

felixschlegel:
> **Note**: this PR sits on top of #87.

> Adds some of the missing properties from #46.

### Motivation:

We were lacking some of the security protocol configuration options and
want to provide them in a type-safe manner.

### Modifications:

* add Kerberos support
* add OAuthBearer support
* Created a new file `KafkaConfiguration+Security.swift`
    * add `public struct` `SSLConfiguration`
    * add `public struct` `SASLMechanism`
    * add `public struct` `KerberosConfiguration`
    * add `public struct` `OAuthBearerMethod`
    * add `public struct` `SecurityProtocol`
* integrate `KafkaConfiguration.SSLOptions` in new types
* integrate `KafkaConfiguration.SASLOptions` in new types

### Example:

```swift
var config = KafkaConsumerConfiguration(
    consumptionStrategy: .partition(.unassigned, topic: "some topic")
)
config.securityProtocol = .saslSSL(
    saslMechanism: .gssapi(kerberosConfiguration: .init(keytab: "/path/to/keytab")),
    sslConfiguaration: .keyPair(
        privateKey: .init(
            location: .pem("jfklaffalk"),
            password: "iLoveApple"
        ),
        publicKeyCertificate: .file(location: "/path/to/file"),
        caCertificate: .pem("jfkaljfl"),
        crlLocation: nil
    )
)
```

### DIFF
--- a/Sources/SwiftKafka/Configuration/KafkaConfiguration+Security.swift
+++ b/Sources/SwiftKafka/Configuration/KafkaConfiguration+Security.swift
@@ -17,8 +17,9 @@ extension KafkaConfiguration {
 
     /// Use to configure an SSL connection.
     public struct SSLConfiguration: Sendable, Hashable {
-        /// A key either located in a file or directly as a key String (PEM format).
-        public struct Key: Sendable, Hashable {
+        /// Certificate chain consisting of one leaf certificate and potenentially multiple intermediate certificates.
+        /// The public key of the leaf certificate will be used for authentication.
+        public struct LeafAndIntermediates: Sendable, Hashable {
             internal enum _Key: Sendable, Hashable {
                 case file(location: String)
                 case pem(String)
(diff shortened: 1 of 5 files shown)


Sources/SwiftKafka/Configuration/KafkaConfiguration+Security.swift
```diff
@@ -0,0 +1,453 @@
...
+        /// - Parameters:
+        ///
+        ///     - privateKey: The client's private key (PEM) used for authentication.
+        ///     - publicKeyCertificate: The client's public key (PEM) used for authentication.
+        ///     - caCertificate: File or directory path to CA certificate(s) for verifying the broker's key.
+        ///     - crlocation: Path to CRL for verifying broker's certificate validity.
+        public static func keyPair(
+            privateKey: PrivateKey,
+            publicKeyCertificate: Key,
+            caCertificate: Key,
```
dnadoba, line 97: I think we should provide an easy way to use the "default" i.e. "probe" that searches for locally trusted CAs.

Sources/SwiftKafka/Configuration/KafkaConfiguration+Security.swift
```diff
@@ -0,0 +1,453 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+extension KafkaConfiguration {
+    // MARK: - SSLConfiguration
+
+    /// Use to configure an SSL connection.
+    public struct SSLConfiguration: Sendable, Hashable {
+        /// A key either located in a file or directly as a key String (PEM format).
+        public struct Key: Sendable, Hashable {
```
dnadoba, line 21: Probably the right term here would be `Chain` as it not only contains the public key but at least one certificate and possibly multiple certificates.

Sources/SwiftKafka/Configuration/KafkaConfiguration+Security.swift
dnadoba, line 97: we might want to use different types here so we can provide a reasonable default for `caCertificate`.

Sources/SwiftKafka/Configuration/KafkaConfiguration+Security.swift
dnadoba: Hm... maybe even two types.
- One of them is kind of a `LeafAndIntermediates` type as the first certificate in the chain will be used as the identity of the client and the remaining, in no particular order, will be used as potential intermediate certificates and send to the broker to verify the client.
- The other type could be `RootCertificates` as it is an unordered list of trusted CAs that is used by the client to verifier the identity of the broker.

FranzBusch (approved):

Sources/SwiftKafka/Configuration/KafkaConfiguration+Security.swift
```diff
@@ -0,0 +1,513 @@
...
+            return resultDict
+        }
+    }
+
+    // MARK: - SASLMechanism
+
+    /// Available SASL mechanisms that can be used for authentication.
+    public struct SASLMechanism: Sendable, Hashable {
+        /// Used to configure Kerberos.
+        public struct KerberosConfiguration: Sendable, Hashable {
```
dnadoba, line 242: I think there is a public init missing

Sources/SwiftKafka/Configuration/KafkaConfiguration+Security.swift
```diff
@@ -0,0 +1,513 @@
...
+
+        private let _internal: _SecurityProtocol
+
+        /// Send messages as plaintext (no security protocol used).
+        public static let plaintext = SecurityProtocol(
+            _internal: .plaintext
+        )
+
+        /// Use the Secure Sockets Layer (SSL) protocol.
+        public static func ssl(configuration: SSLConfiguration) -> SecurityProtocol {
```
dnadoba, line 456: @lukasa should this be called TLS instead of SSL?

dnadoba (commented):
Do we allow users to modify the config dictionary? e.g. to allow users using config options the swift wrappers do not yet support?

dnadoba (approved):
Awesome! Looks good to me

Issue #79, opened 2023-06-29
Redefine statistics callback


mr-swifter:
`librdkafka` provides statistics over a callback which can be set using `rd_kafka_conf_set_stats_cb`.

Statistics callback should be set in combination with `statistics.interval.ms`.

Reference https://docs.confluent.io/platform/current/clients/librdkafka/html/md_STATISTICS.html

https://quicktype.io can be used to generate proper structs for statistics content.

blindspotbounty:
@felixschlegel, @FranzBusch I have made some approach for this improvement.
Unfortunately, I have some doubts regarding ideas in the code, so could you give me a piece of advise in the PR, please?

PR #92, opened 2023-07-18
Feature: expose librdkafka statistics as swift metrics

blindspotbounty:
This is approximate solution for exposing librdkafka statistics callback (https://github.com/swift-server/swift-kafka-gsoc/issues/79)

Main ideas:
1. Specify interval for metrics updates
2. Allow to specify metrics per librdkafka key


felixschlegel (changes requested):
Generally a good start! However, I think we should **not** have a statistics closure and all these new initializers, but rather an `AsyncSequence` as we do with `KafkaConsumer.messages` or the producer acknowledgements. The implementation of statistics should therefore follow a similar pattern. Feel free to reach out to me if you need any more pointers for that!

blindspotbounty:
> Generally a good start! However, I think we should **not** have a statistics closure and all these new initializers, but rather an `AsyncSequence` as we do with `KafkaConsumer.messages` or the producer acknowledgements. The implementation of statistics should therefore follow a similar pattern. Feel free to reach out to me if you need any more pointers for that!

Thank you for the feedback. I think I got you. Basically, we add type to `KafkaEvent`, e.g. `statistics` and use it across the code.
Let me try it out, it sounds promising regarding how it should look like.

felixschlegel:
> > Generally a good start! However, I think we should **not** have a statistics closure and all these new initializers, but rather an `AsyncSequence` as we do with `KafkaConsumer.messages` or the producer acknowledgements. The implementation of statistics should therefore follow a similar pattern. Feel free to reach out to me if you need any more pointers for that!
> 
> Thank you for the feedback. I think I got you. Basically, we add type to `KafkaEvent`, e.g. `statistics` and use it across the code. Let me try it out, it sounds promising regarding how it should look like.

Yes, so you would probably want to add that to `KafkaEvent`, receive the `statistics` event through `KafkaClient.eventPoll` and then `yield` the statistics to some new `AsyncSequence` that is exposed by `KafkaConsumer` + `KafkaProducer`.

blindspotbounty:
> > > Generally a good start! However, I think we should **not** have a statistics closure and all these new initializers, but rather an `AsyncSequence` as we do with `KafkaConsumer.messages` or the producer acknowledgements. The implementation of statistics should therefore follow a similar pattern. Feel free to reach out to me if you need any more pointers for that!
> > 
> > 
> > Thank you for the feedback. I think I got you. Basically, we add type to `KafkaEvent`, e.g. `statistics` and use it across the code. Let me try it out, it sounds promising regarding how it should look like.
> 
> Yes, so you would probably want to add that to `KafkaEvent`, receive the `statistics` event through `KafkaClient.eventPoll` and then `yield` the statistics to some new `AsyncSequence` that is exposed by `KafkaConsumer` + `KafkaProducer`.

I pushed a draft for Consumer (not touching producer so far), just to be sure that I got the idea correctly. If it is okay, I will perform other cleanup in code and put the same code for producer as well.

Tests/SwiftKafkaTests/KafkaConsumerTests.swift
```diff
@@ -85,4 +87,92 @@ final class KafkaConsumerTests: XCTestCase {
...
+        let uniqueGroupID = UUID().uuidString
+        var config = KafkaConsumerConfiguration(
+            consumptionStrategy: .group(id: uniqueGroupID, topics: ["this-topic-does-not-exist"])
+        )
+        config.statisticsInterval = Duration.milliseconds(100)
+
+        let stringJson = NIOLockedValueBox<String>(String())
+        let consumer = try KafkaConsumer(config: config, logger: .kafkaTest)
+
+        let statTask = Task {
```
felixschlegel, line 102: This can be part of the TaskGroup I think

blindspotbounty: Oh, certainly yes, thank you for the hint!
Should we probably collect all tasks from the group as they all should be finished? 
Like this:
```
for try await _ in group { }
```

felixschlegel: You invoke `group.next()` to receive the result of the next completed task. You don't have to invoke it for all of the Tasks you add. When you take a look at the `SwiftKafkaTests.swift` you can see that I used `group.next()` to wait for the producer/consumer task to be finished so I can safely `triggerGracefulShutdown`  _after_ all my producer/consumer testing logic has run.

[Documentation for reference](https://developer.apple.com/documentation/swift/taskgroup/next())

Sources/SwiftKafka/Utilities/KafkaStatistics.swift
```diff
@@ -0,0 +1,70 @@
+import NIOConcurrencyHelpers
+import NIOCore
+
+public typealias KafkaStatistics = String
+
+public struct KafkaStatisticsSeqence: Sendable, AsyncSequence {
+    struct EmptyDelegate: NIOAsyncSequenceProducerDelegate {
```
blindspotbounty, line 7: I was thinking whether we need anything to do with this sequence, i.e.:
1. to call `finish()` in Consumer/Producer delegate or
2. to call it here

felixschlegel: I think ideally what we would do here is something like we do in the `KafkaConsumer` where we enter a state in which we are still polling for events but then in this case drop any incoming statistics events so that new statistics emitted after termination of the `AsyncSequence` don't buffer in memory.

blindspotbounty: The main problem is that there is one state machine, so one switch to `.consumptionStopped` should be enough on terminate of any sequence

felixschlegel: Yeah, we have to work around this here. There are two things that I can currently think of:

1. In `KafkaClient` (now `RDKafkaClient` on `main`) keep track of all the `RDKafkaEvent`s we are listening to and drop inside of `KafkaClient` when `RDKafkaEvent` is not part of the list of `RDKafkaEvent`s we are listening to -> in `KafkaConsumer` / `KafkaProducer` you would then just change what the `RDKafkaClient` should listen to
2. Have something like a `StatisticaHandler` type that has its own state and just forwards stuff to the `source` or not

felixschlegel (changes requested):

Sources/SwiftKafka/Utilities/KafkaStatisticsJsonModel.swift
```diff
@@ -0,0 +1,177 @@
...
+//===----------------------------------------------------------------------===//
+
+// This file was generated from JSON Schema using quicktype, do not modify it directly.
+// To parse the JSON, add this file to your project and do:
+//
+//   let statistics = try? newJSONDecoder().decode(KafkaStatisticsJsonModel.self, from: jsonData)
+
+// MARK: - Statistics
+
+public struct KafkaStatisticsJson: Hashable, Codable {
```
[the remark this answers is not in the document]
blindspotbounty: Thank you for your comment!

I think that `swift-metrics` is a good thing and should be used to deliver statistics and that is the most important thing.

However, from my perspective, there could be other scenarios when statistics can be used by an end-user application directly for adjusting its behaviour. 
There are some examples: 
1. measure lag to the final offset to estimate [when/are/time for] all records received 
2. work with latency metrics to provide approximate latency for end user
3. understand the number messages in flight to adjust internal throttling/coalescing/backpressure

I believe there could be other various scenarios when it is required by end-user application. However..

Next thing that comes to mind is about delivery of this statistics to `swift-metrics`: I am not sure if there is an automatic way to deliver the whole json to swift-metics directly...
Therefore, I am a little bit confused if there is a suggestion to use some other model instead of proposed. 
Should we?

Sources/SwiftKafka/KafkaConsumerEvent.swift
```diff
@@ -14,11 +14,15 @@
 
 /// An enumeration representing events that can be received through the ``KafkaConsumerEvents`` asynchronous sequence.
 public enum KafkaConsumerEvent: Sendable, Hashable {
+    /// Statistics from librdkafka
+    case statistics(KafkaStatistics)
     /// - Important: Always provide a `default` case when switiching over this `enum`.
     case DO_NOT_SWITCH_OVER_THIS_EXHAUSITVELY
 
-    internal init(_ event: RDKafkaClient.KafkaEvent) {
+    internal init?(_ event: RDKafkaClient.KafkaEvent) {
```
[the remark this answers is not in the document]
blindspotbounty: That is definitely not required any more.

Tests/SwiftKafkaTests/KafkaConsumerTests.swift
```diff
@@ -85,4 +86,53 @@ final class KafkaConsumerTests: XCTestCase {
...
+
+        try await withThrowingTaskGroup(of: Void.self) { group in
+            // Run Task
+            group.addTask {
+                try await serviceGroup.run()
+            }
+
+            // check for librdkafka statistics
+            group.addTask {
+                for try await event in events {
```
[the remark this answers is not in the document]
blindspotbounty: That sounds better

Sources/SwiftKafka/Configuration/KafkaConsumerConfiguration.swift
```diff
@@ -23,6 +23,18 @@ public struct KafkaConsumerConfiguration {
     /// Default: `.milliseconds(100)`
     public var pollInterval: Duration = .milliseconds(100)
 
+    /// Interval for librdkafka statistics reports
+    /// 0ms - disabled
+    /// >= 1ms - statistics provided every specified interval
```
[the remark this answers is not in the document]
blindspotbounty: I was thinking about property wrapper but looked at other properties that are checked in place, e.g.:
```
    /// Metadata cache max age.
    /// (Lowest granularity is milliseconds)
    /// Default: `.milliseconds(900_000)`
    public var maximumMetadataAge: Duration = .milliseconds(900_000) {
        didSet {
            precondition(
                self.maximumMetadataAge.canBeRepresentedAsMilliseconds,
                "Lowest granularity is milliseconds"
            )
        }
    }
```

Sources/SwiftKafka/Utilities/KafkaStatisticsJsonModel.swift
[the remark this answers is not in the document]
felixschlegel: Hello @blindspotbounty ,

I don't think providing metrics to an end-user application is a priority. Users should just tweak their configurations to work well and should not have some statistics-dependent magic running in their applications IMO.

Unfortunately, there is no direct mapping from `json` to `swift-metrics`. However, here is how I would go about implementing this:

1. Have a new type called `KafkaMetricsReporting` that you pass as an _optional_ argument as part of `KafkaConsumerConfiguration`/`KafkaProducerConfiguration`

This could look something like this:

```swift
public struct KafkaMetricsReporting {
    public let myCounter: Metrics.Counter?
    public let myGauge: Metrics.Gauge?
    public let myTimer: Metrics.Timer?
}
```

```swift
public struct KafkaConsumerConfig {
    public var metrics: KafkaMetricsReporting?
}
```

This can be similar to the `KafkaStatistics` type you already have. However, you'd have to find suitable [metric types](https://github.com/apple/swift-metrics#metric-types) for each statistics property.

2. If the `Kafka[Consumer|Producer]Configuration` is non-`nil` and not all of its properties are `nil`, enable for statistics and update the metrics accordingly when a new statistics event is received

3. When you have received the event, update all metrics in you `KafkaMetricsReporting` that are non-`nil`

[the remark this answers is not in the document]
blindspotbounty: That sounds nice and should be pretty much convenient.
However, there could be a number of topics and partitions in them. Currently, I see just several solutions that might suit here:
1. Provide metrics factory that would make counters with labels topic=abc partition=xx
2. Calculate averages/totals for all topics/partitions for specific counters (probably the best)
3. Provide different gauges/counters in a dictionary to configuration
4. Don't cover array fields so far
 
Maybe there is some better solution, let me know if you have something in mind, please.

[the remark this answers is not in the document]
felixschlegel: Hey @blindspotbounty , I haven’t really thought about how we should handle these array based metrics for topic / partitions etc. however I am inclined to not aggregate them to averages but rather expose topic / partition … -specific information.

Metrics has evolved into quite a big feature now so I would suggest not covering all metrics at once but rather start by getting the general structure right and listening to a hand full of metrics.

You can then just add the remaining metrics in a follow-up pull request once we get the basics right :smile:

I believe starting with some of the [Top Level Metrics](https://github.com/confluentinc/librdkafka/blob/master/STATISTICS.md#top-level) is a good start

felixschlegel (changes requested):
Great work! Please see my comments 😄

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,65 @@
...
+        public var msgCount: Gauge?
+        /// Current total size of messages in producer queues
+        public var msgSize: Gauge?
+        /// Threshold: maximum number of messages allowed allowed on the producer queues
+        public var msgMax: Gauge?
+        /// Threshold: maximum total size of messages allowed on the producer queues
+        public var msgSizeMax: Gauge?
+        
+        /// Total number of requests sent to Kafka brokers
+        public var tx: Gauge?
```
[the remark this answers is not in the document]
blindspotbounty: I took them from librdkafka statistics. 
So, rename them all?

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,65 @@
...
+        /// Total number of message bytes (including framing) received from Kafka brokers
+        public var rxMessagesBytex: Gauge?
+        
+        /// Number of topics in the metadata cache.
+        public var metadataCacheCount: Gauge?
+    }
+    
+    public enum Metrics: Sendable {
+        case disable
+        case enable(updateInterval: KafkaConfiguration.KeyRefreshAttempts, options: MetricsOptions)
```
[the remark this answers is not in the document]
blindspotbounty: The suggestion was there https://github.com/swift-server/swift-kafka-client/pull/92#discussion_r1286902597
Should we back to Duration?

Sources/Kafka/KafkaProducer.swift
```diff
@@ -208,6 +214,12 @@ public final class KafkaProducer: Service, Sendable {
             case .pollAndYield(let client, let source):
                 let events = client.eventPoll()
                 for event in events {
+                    if case let .statistics(kafkaStatistics) = event {
```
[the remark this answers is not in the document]
blindspotbounty: Sure

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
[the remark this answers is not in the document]
felixschlegel: Yes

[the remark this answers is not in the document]
blindspotbounty: Btw, should we rename them in json model?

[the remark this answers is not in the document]
felixschlegel: Not a must as long as the JSON model remains `internal`, but it would make the code a lot more coherent 😄

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -241,6 +247,9 @@ public final class KafkaConsumer: Sendable, Service {
         if configuration.isAutoCommitEnabled == false {
             subscribedEvents.append(.offsetCommit)
         }
+        if case .enable = configuration.metrics {
```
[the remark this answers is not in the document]
blindspotbounty: I've added `didSet` with precondition for both: enabled metrics and for positive duration.

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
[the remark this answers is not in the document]
blindspotbounty: I can rename those field then mapping will just in coding keys.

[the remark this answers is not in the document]
felixschlegel: Great!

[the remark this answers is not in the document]
blindspotbounty: done

felixschlegel (changes requested):
Hey @blindspotbounty,

Again, excellent work!
I have left some minor comments inline and also a comment regarding how we are "filling in" the statistics.

-Felix

Sources/Kafka/Utilities/KafkaStatistics.swift
```diff
@@ -0,0 +1,65 @@
...
+// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import ExtrasJSON
+import Metrics
+
+struct KafkaStatistics: Sendable, Hashable {
```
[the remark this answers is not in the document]
blindspotbounty: Nice idea!

Sources/Kafka/Utilities/KafkaStatistics.swift
```diff
@@ -0,0 +1,65 @@
...
+            record(json.totalMessagesSent, to: options.totalMessagesSent)
+            record(json.totalBytesSent, to: options.totalBytesSent)
+            
+            record(json.totalMessagesBytesSent, to: options.totalMessagesBytesSent)
+            record(json.totalMessagesRecieved, to: options.totalMessagesRecieved)
+            record(json.totalMessagesBytesRecieved, to: options.totalMessagesBytesRecieved)
+            record(json.metadataCacheCount, to: options.metadataCacheCount)
+
+        } catch {
+            fatalError("Statistics json decode error \(error)")
```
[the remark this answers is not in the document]
blindspotbounty: Let's make an assert?

Package.swift
```diff
@@ -48,9 +48,11 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.55.0"),
         .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.0.0-beta.1"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-metrics", from: "2.4.1"),
         // The zstd Swift package produces warnings that we cannot resolve:
         // https://github.com/facebook/zstd/issues/3328
         .package(url: "https://github.com/facebook/zstd.git", from: "1.5.0"),
+        .package(url: "https://github.com/swift-extras/swift-extras-json.git", .upToNextMajor(from: "0.6.0")),
```
FranzBusch, line 55: Can we please use `Foundation.JSONDecoder` here instead.

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,133 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Metrics
+
+extension KafkaConfiguration {
+    // MARK: - Metrics
+
+    /// Use to configure metrics.
+    public struct KafkaMetrics: Sendable {
```
FranzBusch, line 21:
```suggestion
    public struct Metrics: Sendable {
```

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,133 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Metrics
+
+extension KafkaConfiguration {
+    // MARK: - Metrics
+
+    /// Use to configure metrics.
```
FranzBusch, line 20:
```suggestion
    /// Configuration for the metrics emitted by `SwiftKafka`.
```

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,133 @@
...
+                self.totalBytesReceived != nil ||
+                self.metadataCacheCount != nil
+        }
+
+        /// librdkafka's internal monotonic clock (microseconds)
+        public var timestamp: Gauge?
+        /// Wall clock time in seconds since the epoch
+        public var time: Timer?
+        /// Time since this client instance was created
+        public var age: Timer?
```
FranzBusch, line 46: I don't know if we need those. Do we have a concrete use-case how those would be observed?

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,133 @@
...
+        /// Number of ops (callbacks, events, etc) waiting in queue for application to serve
+        public var replyQueue: Gauge?
```
FranzBusch, line 48:
```suggestion
        /// Number of operations (callbacks, events, etc) waiting in the queue.
        public var queuedOperation: Gauge?
```

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,133 @@
...
+        /// Current number of messages in producer queues
+        public var messageCount: Gauge?
+        /// Current total size of messages in producer queues
+        public var messageSize: Gauge?
+        /// Threshold: maximum number of messages allowed allowed on the producer queues
+        public var messageMax: Gauge?
+        /// Threshold: maximum total size of messages allowed on the producer queues
+        public var messageSizeMax: Gauge?
```
FranzBusch, line 56: Isn't this threshold set by the configuration? Can this change over time? Wondering if should expose this

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,133 @@
...
+        /// Wall clock time in seconds since the epoch
+        public var time: Timer?
+        /// Time since this client instance was created
+        public var age: Timer?
+        /// Number of ops (callbacks, events, etc) waiting in queue for application to serve
+        public var replyQueue: Gauge?
+        /// Current number of messages in producer queues
+        public var messageCount: Gauge?
+        /// Current total size of messages in producer queues
+        public var messageSize: Gauge?
```
FranzBusch, line 52: This is producer specific should we only make this available on the producer configuration?

```suggestion
        /// Current number of queued producer messages.
        public var queuedProducerMessages: Gauge?
        /// Current total size in bytes of queued producer messages.
        public var queuedProducerMessagesSize: Gauge?
```

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,133 @@
...
+        /// Threshold: maximum number of messages allowed allowed on the producer queues
+        public var messageMax: Gauge?
+        /// Threshold: maximum total size of messages allowed on the producer queues
+        public var messageSizeMax: Gauge?
+
+        /// Total number of requests sent to Kafka brokers
+        public var totalRequestsSent: Gauge?
```
FranzBusch, line 59: If this is a total then it shouldn't be a `Gauge` but rather a `Counter` or are we using a `Gauge` because we only get the latest values and just have to set it?
```suggestion
        /// Total number of requests sent to Kafka brokers.
        public var totalKafkaBrokerRequests: Counter?
```

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,133 @@
...
+        /// Total number of bytes transmitted to Kafka brokers
+        public var totalBytesSent: Gauge?
+        /// Total number of responses received from Kafka brokers
+        public var totalResponsesRecieved: Gauge?
+        /// Total number of bytes received from Kafka brokers
+        public var totalBytesReceived: Gauge?
```
FranzBusch, line 65:
```suggestion
        /// Total number of bytes transmitted to Kafka brokers.
        public var totalKafkaBrokerRequestsSize: Counter?
        /// Total number of responses received from Kafka brokers.
        public var totalKafkaBrokerResponses: Counter?
        /// Total number of bytes received from Kafka brokers.
        public var totalKafkaBrokerResponsesSize: Gauge?
```

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,133 @@
...
+
+        /// Total number of messages transmitted (produced) to Kafka brokers
```
FranzBusch, line 67: Same for all of those

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,133 @@
...
+        public var totalMessagesSent: Gauge?
+        /// Total number of message bytes (including framing, such as per-Message framing and MessageSet/batch framing) transmitted to Kafka brokers
+        public var totalMessagesBytesSent: Gauge?
+        /// Total number of messages consumed, not including ignored messages (due to offset, etc), from Kafka brokers.
+        public var totalMessagesRecieved: Gauge?
+        /// Total number of message bytes (including framing) received from Kafka brokers
+        public var totalMessagesBytesRecieved: Gauge?
+
+        /// Number of topics in the metadata cache.
+        public var metadataCacheCount: Gauge?
```
FranzBusch, line 77:
```suggestion
        /// Number of topics in the metadata cache.
        public var topicsInMetadataCache: Gauge?
```

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,133 @@
...
+            Self.record(rdKafkaStatistics.totalBytesSent, to: self.totalBytesSent)
+
+            Self.record(rdKafkaStatistics.totalMessagesBytesSent, to: self.totalMessagesBytesSent)
+            Self.record(rdKafkaStatistics.totalMessagesRecieved, to: self.totalMessagesRecieved)
+            Self.record(rdKafkaStatistics.totalMessagesBytesRecieved, to: self.totalMessagesBytesRecieved)
+            Self.record(rdKafkaStatistics.metadataCacheCount, to: self.metadataCacheCount)
+        }
+    }
+
+    public enum Metrics: Sendable {
```
FranzBusch, line 119: We don't nee this struct. Let's just make it an optional in the producer and consumer configuration and add the `updateInterval` into the current `KafkaMetrics` struct.

Sources/Kafka/RDKafka/RDKafkaStatistics.swift
```diff
@@ -0,0 +1,184 @@
...
+
+// This file was generated from JSON Schema using quicktype, do not modify it directly.
+// To parse the JSON, add this file to your project and do:
+//
+//   let statistics = try? newJSONDecoder().decode(RDKafkaStatistics.self, from: jsonData)
+
+// MARK: - Statistics
+
+struct RDKafkaStatistics: Hashable, Codable {
+    let name, clientID, type: String?
```
FranzBusch, line 23: Can we please put each property into a separate line here

Sources/Kafka/RDKafka/RDKafkaStatistics.swift
```diff
@@ -0,0 +1,184 @@
...
+    let timestamp, time, age, replyQueue: Int?
+    let messageCount, messageSize, messageMax, messageSizeMax: Int?
+    let simpleCnt, metadataCacheCount: Int?
+//    let brokers: [String: Broker]?
+//    let topics: [String: Topic]?
+//    let cgrp: Cgrp?
```
FranzBusch, line 29: Can we remove the commented out code please

Tests/KafkaTests/KafkaConsumerTests.swift
```diff
@@ -82,4 +83,43 @@ final class KafkaConsumerTests: XCTestCase {
...
+    func testConsumerStatistics() async throws {
+        let uniqueGroupID = UUID().uuidString
+        var config = KafkaConsumerConfiguration(
+            consumptionStrategy: .group(id: uniqueGroupID, topics: ["this-topic-does-not-exist"]),
+            bootstrapBrokerAddresses: []
+        )
+
+        var metricsOptions = KafkaConfiguration.KafkaMetrics()
+
+        let handler = MockTimerHandler()
```
FranzBusch, line 96: There is a `MetricsTestingKit` which we should use here instead so we don't need the `MockTimerHandler`

FranzBusch (changes requested):
Overall looks good and thanks for the work. I left some comments inline

Package.swift
blindspotbounty: That make sense. As seems that JSONDecoder is supported for all platforms
Thank you

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
blindspotbounty: I don't see much value in timestamp and time. However, age of specific client can be useful to see its correlation with sent/received data and other things.

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
blindspotbounty: That is good question. I will check and remove if it is static fields

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
blindspotbounty: Yep, I think we can

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
blindspotbounty: I agree that it should be a counter but your guess is right - we receive the latest values for this field. Therefore, using gauge here.

Sources/Kafka/RDKafka/RDKafkaStatistics.swift
blindspotbounty: Sure!

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
blindspotbounty: Let's do that way

Sources/Kafka/RDKafka/RDKafkaStatistics.swift
blindspotbounty: Let's remove it for now and add back when will add those metrics

Tests/KafkaTests/KafkaConsumerTests.swift
blindspotbounty: Good to know, let me use it

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
FranzBusch: Can we leave the age out for now as well. This is always something that can be tracked outside of kafka itself

FranzBusch:
@blindspotbounty If you are ready and can push up the changes addressing the review, I am happy to get this over the line

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
blindspotbounty: I've removed them as they are static

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
blindspotbounty: I've renamed it but do you think we should make `ProducerMetrics` and `ConsumerMetrics` and allow only specific ones?

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
blindspotbounty: I've renamed all of them, let me know if you have some other names in mind

blindspotbounty:
@FranzBusch I've pushed commit with addressing conversations. But remained two conversations open in case it should be different.

Tests/KafkaTests/Utilities.swift
```diff
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import Logging
+import Metrics
```
felixschlegel, line 16: Do we still need this here?

blindspotbounty: Surely not! Thank you for catching that :)

Sources/Kafka/RDKafka/RDKafkaStatistics.swift
```diff
@@ -0,0 +1,50 @@
...
+// See CONTRIBUTORS.txt for the list of swift-kafka-client project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+// This file was generated from JSON Schema using quicktype, do not modify it directly.
+// To parse the JSON, add this file to your project and do:
+//
+//   let statistics = try? newJSONDecoder().decode(RDKafkaStatistics.self, from: jsonData)
```
FranzBusch, line 18: Can we remove this comment. I assume we are going to hand edit it in the future.

Sources/Kafka/Configuration/KafkaConfiguration+Metrics.swift
```diff
@@ -0,0 +1,94 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Metrics
+
+extension KafkaConfiguration {
+    // MARK: - Metrics
+
+    /// Configuration for the metrics emitted by `SwiftKafka`.
+    public struct Metrics: Sendable {
```
FranzBusch, line 21: I think this looks really good now. The only thing left is that I think we should split this into two structs on inside the producer and one inside the consumer configuration and only pull out the fields that make sense to them. Then we should be good to merge this.

FranzBusch (commented):
I think we are really close! Thanks for all the work on this one last comment

Sources/Kafka/RDKafka/RDKafkaStatistics.swift
blindspotbounty: removed!

FranzBusch (approved):
Looks really good! Thanks for doing the work.

FranzBusch:
@swift-server-bot add to allowlist

FranzBusch:
@swift-ci please test

blindspotbounty:
@FranzBusch sure. But I think it is better to restart ci after resolving conflicts.

FranzBusch:
@swift-ci please test

blindspotbounty:
Seems, it is the same as for PR for backpressure:
```
11:12:15 /code/Sources/Kafka/RDKafka/RDKafkaClient.swift:33:17: error: stored property 'kafkaHandle' of 'Sendable'-conforming class 'RDKafkaClient' has non-sendable type 'OpaquePointer'
```
https://github.com/swift-server/swift-kafka-client/pull/139#issuecomment-1772685064

PR #94, opened 2023-07-20
`KafkaProducer`: refactor flushing

felixschlegel:
### Motivation:

With our own flushing implementation we were flushing until `rd_kakfa_outq_len`
reached `0`. However `rd_kakfa_outq_len` also takes other events such
as statistics into account which led to a race where we were flushing
for a very long time because more and more other events were produced to
the queue while we were flushing.

### Modifications:

* `KafkaClient`:
    * remove `var outgoingQueueSize`
    * add new method `flush(timeoutMilliseconds:)` that executes the
      blocking `rd_kafka_flush` call on a `DispatchQueue` but vends this
      as an `async func`
* `KafkaProducer`:
    * rename `KafkaProducer.StateMachine.State.flushing` to
      `KafkaProducer.StateMachine.State.finishing`
    * invoke `KafkaClient.flush` before terminating poll loop

### DIFF
--- a/Sources/SwiftKafka/Configuration/KafkaProducerConfiguration.swift
+++ b/Sources/SwiftKafka/Configuration/KafkaProducerConfiguration.swift
@@ -20,6 +20,10 @@ public struct KafkaProducerConfiguration {
     /// Default: `.milliseconds(100)`
     public var pollInterval: Duration = .milliseconds(100)
 
+    /// Maximum timeout for flushing outstanding produce requests when the ``KakfaProducer`` is shutting down.
+    /// Default: `10000`
+    public var flushTimeoutMilliseconds: Int32 = 10000
+
     // MARK: - Producer-specific Config Properties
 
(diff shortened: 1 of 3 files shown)


felixschlegel:
@FranzBusch : how shall we go about vending the `flush` `timeout` to the user?

FranzBusch:
> how shall we go about vending the flush timeout to the user?

We should put it into our configuration struct.

Sources/SwiftKafka/Configuration/KafkaProducerConfiguration.swift
```diff
@@ -20,6 +20,14 @@ public struct KafkaProducerConfiguration {
     /// Default: `.milliseconds(100)`
     public var pollInterval: Duration = .milliseconds(100)
 
+    /// Maximum timeout for flushing outstanding produce requests when the ``KakfaProducer`` is shutting down.
+    /// Default: `10000`
+    public var flushTimeoutMilliseconds: Int = 10000 {
+        didSet {
+            precondition(0...Int(Int32.max) ~= self.flushTimeoutMilliseconds)
```
FranzBusch, line 27: Can we provide a message here as well when we fail the precondition please

FranzBusch (approved):

PR #95, opened 2023-07-21
`KafkaProducerMessage`: generic key & value

felixschlegel:
### Motivation:

We want to allow our users to feed arbitrary data as `key` and `value`
of a `KafkaProducerMessage`.

### Modifications:

* add new `public protocol KafkaBuffer` which asks implementing types to
  provide safe access to an `UnsafeMutableRawBufferPointer` containing
  the data
* implement `KafkaBuffer` in `String`
* implement `KafkaBuffer` in `ByteBuffer`
* update tests


Tests/SwiftKafkaTests/KafkaProducerTests.swift
```diff
@@ -110,7 +110,7 @@ final class KafkaProducerTests: XCTestCase {
             }
 
             let expectedTopic = "test-topic"
-            let message = KafkaProducerMessage(
+            let message = KafkaProducerMessage<ByteBuffer, ByteBuffer>(
```
felixschlegel, line 113: Keys are optional, which leads to the problem that their type cannot be inferred when set to `nil`. This is a workaround but far from perfect. Any suggestions?

Sources/SwiftKafka/Data/String+KafkaBuffer.swift
```diff
@@ -0,0 +1,30 @@
...
+import NIOCore
+
+extension String: KafkaBuffer {
+    /// Yields a buffer pointer containing this `KafkaBuffer`'s readable bytes.
+    ///
+    /// - warning: Do not escape the pointer from the closure for later use.
+    ///
+    /// - parameters:
+    ///     - body: The closure that will accept the yielded bytes.
+    /// - returns: The value returned by `body`.
```
FranzBusch, line 24: Just having this doc on the protocol should be enough.

Sources/SwiftKafka/Data/String+KafkaBuffer.swift
```diff
@@ -0,0 +1,30 @@
...
+    public func withUnsafeRawBufferPointer<T>(_ body: (UnsafeRawBufferPointer) throws -> T) rethrows -> T {
+        try ByteBuffer(string: self).withUnsafeReadableBytes { UnsafeRawBufferPointer in
+            try body(UnsafeRawBufferPointer)
+        }
```
FranzBusch, line 28: We should try to use `self.utf8.withContiguousStorageIfAvailable` and only fallback to bridging into `ByteBuffer` in the slow path

Sources/SwiftKafka/Data/ByteBuffer+KafkaBuffer.swift
```diff
@@ -0,0 +1,30 @@
...
+import NIOCore
+
+extension ByteBuffer: KafkaBuffer {
+    /// Yields a buffer pointer containing this `KafkaBuffer`'s readable bytes.
+    ///
+    /// - warning: Do not escape the pointer from the closure for later use.
+    ///
+    /// - parameters:
+    ///     - body: The closure that will accept the yielded bytes.
+    /// - returns: The value returned by `body`.
```
FranzBusch, line 24: Same here

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -252,7 +252,7 @@ public final class KafkaProducer: Service, Sendable {
     /// of the corresponding ``KafkaAcknowledgedMessage``.
     /// - Throws: A ``KafkaError`` if sending the message failed.
     @discardableResult
-    public func send(_ message: KafkaProducerMessage) throws -> KafkaProducerMessageID {
+    public func send<K, V>(_ message: KafkaProducerMessage<K, V>) throws -> KafkaProducerMessageID {
```
FranzBusch, line 255: Can we name those generic types `Key` and `Value` please

Sources/SwiftKafka/RDKafka/RDKafkaClient.swift
```diff
@@ -97,33 +97,45 @@ final class RDKafkaClient: Sendable {
     /// - Parameter newMessageID: ID that was assigned to the `message`.
     /// - Parameter topicConfig: The ``KafkaTopicConfiguration`` used for newly created topics.
     /// - Parameter topicHandles: Topic handles that this client uses to produce new messages
-    func produce(
-        message: KafkaProducerMessage,
+    func produce<K, V>(
```
FranzBusch, line 100: Same here

Tests/SwiftKafkaTests/KafkaProducerTests.swift
FranzBusch: Can we conform `Never` to `KafkaBuffer` and provide an init that doesn't take a key when we are `Never`.

Sources/SwiftKafka/Data/KafkaBuffer.swift
```diff
@@ -0,0 +1,25 @@
...
+/// A protocol representing a buffer that provides access to its raw memory representation.
+public protocol KafkaBuffer {
+    /// Yields a buffer pointer containing this `KafkaBuffer`'s readable bytes.
+    ///
+    /// - warning: Do not escape the pointer from the closure for later use.
+    ///
+    /// - parameters:
+    ///     - body: The closure that will accept the yielded bytes.
+    /// - returns: The value returned by `body`.
+    func withUnsafeRawBufferPointer<T>(_ body: (UnsafeRawBufferPointer) throws -> T) rethrows -> T
```
glbrntt, line 24: Does this need to have such a low-level requirement? It feels like an API smell to me that the only requirement is an unsafe API.

Is it possible, for example, to rephrase the API in terms of e.g. `RandomAccessCollection` of `UInt8` and have encoders/decoders?

FranzBusch: I was thinking about this as well but I am unsure this fulfils the goals that we want to achieve here. We want to make it possible to have both `key` and especially `value` be able to pass copy-free to `rdkafka`. If we take a collection of UInt8 then I am not sure if we can achieve that.

glbrntt: You do have `withContiguousStorageIfAvailable` for RAC which should fulfil that _most_ of the time (I guess only custom RACs _might_ not implement it).

FranzBusch: Yeah you are right we can probable be generic over `RAC` here instead of creating our own type. In the slow path we just write the `RAC` into byte buffer. @felixschlegel can you try this out.

felixschlegel: I agree that having the extra `KafkaBuffer` type is not ideal. However, I do not find the `RandomAccessCollection` approach more ergonomic (from what I have investigated):

1. Our `KafkaProducerMessage` type's `String` initializer would have to take `String` as parameter value, but stores a `ContiuousArray<UInt8>` which still requires copying

<details>
<summary>See implementation</summary>

```swift
/// Message that is sent by the `KafkaProducer`
public struct KafkaProducerMessage<Key: RandomAccessCollection<UInt8>, Value: RandomAccessCollection<UInt8>> {
    public var topic: String
    public var partition: KafkaPartition
    public var key: Key?
    public var value: Value

    /// Create a new `KafkaProducerMessage` with a key and value.
    /// - Parameter topic: The topic the message will be sent to. Topics may be created by the `KafkaProducer` if non-existent.
    /// - Parameter partition: The topic partition the message will be sent to. If not set explicitly, the partiotion will be assigned automatically.
    /// - Parameter key: Used to guarantee that messages with the same key will be sent to the same partittion so that their order is preserved.
    /// - Parameter value: The message body.
    public init(
        topic: String,
        partition: KafkaPartition? = nil,
        key: Key,
        value: Value
    ) {
        self.topic = topic
        self.key = key
        self.value = value

        if let partition = partition {
            self.partition = partition
        } else {
            self.partition = .unassigned
        }
    }
}

// MARK: - KafkaProducerMessage + String Initializer

extension KafkaProducerMessage where Key == ContiguousArray<UInt8>, Value == ContiguousArray<UInt8> {

    /// Create a new `KafkaProducerMessage` with a key and value.
    /// - Parameter topic: The topic the message will be sent to. Topics may be created by the `KafkaProducer` if non-existent.
    /// - Parameter partition: The topic partition the message will be sent to. If not set explicitly, the partiotion will be assigned automatically.
    /// - Parameter key: Used to guarantee that messages with the same key will be sent to the same partittion so that their order is preserved.
    /// - Parameter value: The message body.
    public init(
        topic: String,
        partition: KafkaPartition? = nil,
        key: String? = nil,
        value: String
    ) {
        self.topic = topic
        self.key = key.map { ContiguousArray($0.utf8) }
        self.value = ContiguousArray(value.utf8)

        if let partition = partition {
            self.partition = partition
        } else {
            self.partition = .unassigned
        }
    }
}
```

</details>

2. We would have to take `ByteBufferView` as parameter type instead of `ByteBuffer` as only `ByteBufferView` conforms to `RandomAccessCollection`

So yes it is a smell, but we provide default implementations of `KafkaBuffer` for `String` and `ByteBuffer`. The protocol gives the user the freedom to make their own types conform to `KafkaBuffer` at their own risk.  Furthermore this enables them to avoid copying their own types.

glbrntt: Ah right, `UTF8View` is not a `RAC`, but it is a `BidirectionalCollection` which should be sufficient to avoid the copy. I don't think having convenience initialisers (for `String` and `ByteBuffer`) is an issue either. I'd rather reuse existing protocols if possible but don't feel too strongly.

Sources/SwiftKafka/RDKafka/RDKafkaClient.swift
```diff
@@ -97,33 +97,45 @@ final class RDKafkaClient: Sendable {
...
-                    topicHandle,
-                    message.partition.rawValue,
-                    RD_KAFKA_MSG_F_COPY,
-                    UnsafeMutableRawPointer(mutating: valueBuffer.baseAddress),
-                    valueBuffer.count,
-                    keyBytes,
-                    keyBytes?.count ?? 0,
-                    UnsafeMutableRawPointer(bitPattern: newMessageID)
-                )
+        let responseCode = try message.value.withUnsafeRawBufferPointer { valueBuffer in
```
blindspotbounty, line 106: Can we create method in message?
```
func withUnsafeKeyValueRawBufferPointer(_ closure: (Buffer?, Buffer)) {
    let responseCode = try value.withUnsafeRawBufferPointer { valueBuffer in
        if let key {
            key.withUnsafeRawBufferPointer { keyBuffer in
                closure(keyBuffer, valueBuffer)
            }
        } else {
            closure(nil, valueBuffer)
        }
    }
}
```
Then we could avoid call `key.withUnsafeRawBufferPointer` for every topic. Especially, it might make sense for custom types.

blindspotbounty: Ou, sorry, I see that it is only for one topic.

blindspotbounty (approved):

Sources/SwiftKafka/Data/String+KafkaContiguousBytes.swift
```diff
@@ -0,0 +1,32 @@
...
+            // Fast Path
+            let unsafeRawBufferPointer = UnsafeRawBufferPointer(start: unsafePointer.baseAddress, count: self.count)
+            return try body(unsafeRawBufferPointer)
+        }) {
+            return read
+        } else {
+            // Slow path
+            return try ByteBuffer(string: self).withUnsafeReadableBytes { unsafeRawBufferPointer in
+                try body(unsafeRawBufferPointer)
+            }
```
FranzBusch, line 29: NIT: We can just call our other conformance here
```suggestion
            return try ByteBuffer(string: self).withUnsafeBytes(body)
```

Sources/SwiftKafka/KafkaProducerMessage.swift
```diff
@@ -43,30 +43,26 @@ public struct KafkaProducerMessage {
             self.partition = .unassigned
         }
     }
+}
 
-    /// Create a new `KafkaProducerMessage` with a `String` key and value
+extension KafkaProducerMessage where Key == Never {
+    /// Create a new `KafkaProducerMessage` with a `ByteBuffer` key and value
```
FranzBusch, line 49: This looks outdated

FranzBusch (approved):

PR #96, opened 2023-07-24
`KafkaProducer`: events- instead of acks sequence

felixschlegel:
### Motivation:

`KafkaProducer`: we want to expose a general `KafkaProducerEvent` type
in the `AsyncSequence` that was formerly just for message
acknowledgements. This enables us to add more events such as statistics
in the future.

The reason behind having a single `AsyncSequence` for all event types
instead of having separate `AsyncSequence`s for each event type is that
we need to ensure that all events that we subscribe to in `librdkafka`
are actually consumed by the `AsyncSequence` in our `KafkaProducer`
type. Otherwise we could run out of memory.

Now by giving the user the entire events `AsyncSequence`, they decide if
they want to consume event or drop it.

> **Note**: Logs will be consumed regardless of the
> `KafkaProducerEvents` `AsyncSequence`

### Modifications:

* create a new `enum` `KafkaProducerEvent`
* rename `KafkaMessageAcknowledgements` -> `KafkaProducerEvents`
* rename `KafkaProducer.makeProducerWithAcknowledgements` -> `.makeProducerWithEvents`
* update tests
* update README


Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -208,19 +208,16 @@ public final class KafkaProducer: Service, Sendable {
...
+                // Drop any incoming events
                 let _ = client.eventPoll()
             case .pollAndYield(let client, let source):
                 let events = client.eventPoll()
-                for event in events {
-                    switch event {
-                    case .deliveryReport(let results):
+                events
+                    .map(KafkaProducerEvent.fromKafkaEvent(_:))
+                    .forEach {
```
FranzBusch, line 217: Can we keep the for loop here please and just do a switch. The problem with your code is that we actually iterate the events twice. We could use `lazy` but that sometimes fails to optimise so let's rather spell it out.

Sources/SwiftKafka/KafkaProducerEvent.swift
```diff
@@ -0,0 +1,34 @@
...
+
+    /// Down-cast ``RDKafkaClient/KafkaEvent`` to ``KafkaProducerEvent``.
+    static func fromKafkaEvent(_ event: RDKafkaClient.KafkaEvent) -> KafkaProducerEvent {
+        switch event {
+        case .deliveryReport(results: let results):
+            return .deliveryReport(results: results)
+        case .consumerMessages:
+            fatalError("Cannot cast \(event) to KafkaProducerEvent")
+        }
+    }
```
FranzBusch, line 33: This is not really a cast. Can we use an `init(event: RDKafkaClient.KafkaEvent` instead.

Sources/SwiftKafka/KafkaProducerEvent.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+/// An enumeration representing events that can be received through the ``KafkaProducerEvents`` asynchronous sequence.
+public enum KafkaProducerEvent: Hashable, Sendable {
+    /// A delivery report received from the Kafka cluster indicating the status of a produced message.
+    ///
+    /// Parameters:
+    ///    - results: Array of message acknowledgement results.
+    case deliveryReport(results: [Result<KafkaAcknowledgedMessage, KafkaAcknowledgedMessageError>])
```
FranzBusch, line 21: Plural and singular is mixed here
```suggestion
    /// Delivery reports received from the Kafka cluster indicating the status of a produced messages.
    case deliveryReports([Result<KafkaAcknowledgedMessage, KafkaAcknowledgedMessageError>])
```

With this we should probably introduce a new `KafkaDeliveryReport` that is basically a glorified `Result` just with a better name.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -141,28 +141,28 @@ public final class KafkaProducer: Service, Sendable {
...
+    /// - Returns: A tuple containing the created ``KafkaProducer`` and the ``KafkaProducerEvents``
+    /// `AsyncSequence` used for receiving message events.
     /// - Throws: A ``KafkaError`` if initializing the producer failed.
-    public static func makeProducerWithAcknowledgements(
+    public static func makeProducerWithEvents(
         config: KafkaProducerConfiguration = KafkaProducerConfiguration(),
         topicConfig: KafkaTopicConfiguration = KafkaTopicConfiguration(),
         logger: Logger
-    ) throws -> (KafkaProducer, KafkaMessageAcknowledgements) {
+    ) throws -> (KafkaProducer, KafkaProducerEvents) {
```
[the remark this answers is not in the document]
felixschlegel: The consumer will be the same, though I will create a separate PR for this today to keep the PRs smaller to make code review easier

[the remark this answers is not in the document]
felixschlegel: #97

Sources/SwiftKafka/KafkaDeliveryReport.swift
```diff
@@ -0,0 +1,51 @@
...
+    /// The same identifier is returned by ``KafkaProducer/send(_:)`` and can be used to correlate
+    /// a sent message and an acknowledged message.
+    public var id: KafkaProducerMessageID
+
+    internal init?(messagePointer: UnsafePointer<rd_kafka_message_t>?) {
+        guard let messagePointer else {
+            return nil
+        }
+
+        self.id = KafkaProducerMessageID(rawValue: UInt(bitPattern: messagePointer.pointee._private))
```
gjcairo, line 39: Hm, is this access to `_private` safe? It sounds like something we wouldn't want to touch. From [the docs](https://docs.confluent.io/platform/current/clients/librdkafka/html/structrd__kafka__message__t.html#a8eb860a90c978b15892eef54953d5289) it's not really even clear to me what it is that it's referencing.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -41,26 +41,26 @@ extension KafkaProducerCloseOnTerminate: NIOAsyncSequenceProducerDelegate {
...
+public struct KafkaProducerEvents: AsyncSequence {
+    public typealias Element = KafkaProducerEvent
     typealias BackPressureStrategy = NIOAsyncSequenceProducerBackPressureStrategies.NoBackPressure
     typealias WrappedSequence = NIOAsyncSequenceProducer<Element, BackPressureStrategy, KafkaProducerCloseOnTerminate>
     let wrappedSequence: WrappedSequence
 
-    /// `AsynceIteratorProtocol` implementation for handling messages acknowledged by the Kafka cluster (``KafkaAcknowledgedMessage``).
-    public struct AcknowledgedMessagesAsyncIterator: AsyncIteratorProtocol {
+    /// `AsynceIteratorProtocol` implementation for handling ``KafkaProducerEvent``s emitted by Kafka.
+    public struct KafkaProducerEventsAsyncIterator: AsyncIteratorProtocol {
```
gjcairo, line 54: I'm not sure what conventions we're using for naming here, so you can ignore this comment if it goes against what we've been doing elsewhere in this project, but since this struct is nested inside `KafkaProducerEvents` already, I would skip that prefix and name this `AsyncIterator`.
`KafkaProducerEvents.AsyncIterator` seems cleaner than `KafkaProducerEvents.KafkaProducerEventsAsyncIterator`.

felixschlegel:

Sources/SwiftKafka/KafkaDeliveryReport.swift
```diff
@@ -0,0 +1,51 @@
...
+/// A delivery report for a message that was sent to the Kafka cluster.
+public struct KafkaDeliveryReport: Sendable, Hashable {
+    public enum Status: Sendable, Hashable {
+        /// The message has been successfully acknowledged by the Kafka cluster.
+        case acknowledged(message: KafkaAcknowledgedMessage)
+        /// The message failed to be acknowledged by the Kafka cluster and encountered an error.
+        case failure(KafkaError)
+    }
+
+    /// The Status of a Kafka producer message after attempting to send it.
```
FranzBusch, line 26:
```suggestion
    /// The status of a Kafka producer message after attempting to send it.
```

FranzBusch (approved):

gjcairo (approved):

FranzBusch (approved):

PR #97, opened 2023-07-24
`KafkaConsumer`: expose event sequence

felixschlegel:
> **Info**: this PR sits on top of #96

### Motivation:

Like in `KafkaProducer` (#96), we want to expose an asynchronous
sequence that is able to emit all sorts of events in the future (e.g.
rebalance events for the `KafkaConsumer`.

### Modifications:

* add a new type `KafkaConsumerEvent` (currently empty)
* make `KafkaConsumer.init` private
* add factory methods `KafkaConsumer.createConsumer` and
  `KafkaConsumer.createConsumerWithEvents`
* create a new `AsyncSequence`: `KafkaConsumerEvents`
* update README
* update tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ var config = KafkaConsumerConfiguration(
 )
 config.bootstrapServers = [broker]
 
-let consumer = try KafkaConsumer(
+let consumer = try KafkaConsumer.makeConsumer(
     config: config,
     logger: logger
 )
@@ -125,7 +125,7 @@ var config = KafkaConsumerConfiguration(
 )
 config.bootstrapServers = [broker]
 
-let consumer = try KafkaConsumer(
+let consumer = try KafkaConsumer.makeConsumer(
     config: config,
     logger: logger
 )
@@ -163,7 +163,7 @@ var config = KafkaConsumerConfiguration(
 config.enableAutoCommit = false,
 config.bootstrapServers = [broker]
 
-let consumer = try KafkaConsumer(
+let consumer = try KafkaConsumer.makeConsumer(
     config: config,
     logger: logger
 )
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ var config = KafkaConsumerConfiguration(
 )
 config.bootstrapServers = [broker]
 
-let consumer = try KafkaConsumer.makeConsumer(
+let consumer = try KafkaConsumer(
     config: config,
     logger: logger
 )
@@ -125,7 +125,7 @@ var config = KafkaConsumerConfiguration(
 )
 config.bootstrapServers = [broker]
 
-let consumer = try KafkaConsumer.makeConsumer(
+let consumer = try KafkaConsumer(
     config: config,
     logger: logger
 )
@@ -163,7 +163,7 @@ var config = KafkaConsumerConfiguration(
 config.enableAutoCommit = false,
 config.bootstrapServers = [broker]
 
-let consumer = try KafkaConsumer.makeConsumer(
+let consumer = try KafkaConsumer(
     config: config,
     logger: logger
 )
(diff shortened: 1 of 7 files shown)


Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -108,28 +131,25 @@ public final class KafkaConsumer: Sendable, Service {
...
+    // Private initializer, use factory methods to create KafkaConsumer
     /// Initialize a new ``KafkaConsumer``.
     /// To listen to incoming messages, please subscribe to a list of topics using ``subscribe(topics:)``
     /// or assign the consumer to a particular topic + partition pair using ``assign(topic:partition:offset:)``.
-    /// - Parameter config: The ``KafkaConsumerConfiguration`` for configuring the ``KafkaConsumer``.
-    /// - Parameter logger: A logger.
+    ///
+    /// - Parameters:
+    ///     - config: The ``KafkaConsumerConfiguration`` for configuring the ``KafkaConsumer``.
+    ///     - logger: A logger.
```
gjcairo, line 141: `client` and `stateMachine` are missing

gjcairo (approved):
LGTM, just a small comment

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -35,6 +35,29 @@ extension KafkaConsumerCloseOnTerminate: NIOAsyncSequenceProducerDelegate {
     }
 }
 
+// MARK: - KafkaConsumerEvents
+
+/// `AsyncSequence` implementation for handling ``KafkaConsumerEvent``s emitted by Kafka.
+public struct KafkaConsumerEvents: AsyncSequence {
```
FranzBusch, line 41: We should make this `Sendable`

README.md
```diff
@@ -88,7 +88,7 @@ var config = KafkaConsumerConfiguration(
 )
 config.bootstrapServers = [broker]
 
-let consumer = try KafkaConsumer(
+let consumer = try KafkaConsumer.makeConsumer(
```
FranzBusch, line 91: Overall, I still think we should make the `init()` public so that in the case where users are not interested in `event` they can just call `KafkaProducer()` and `KafkaConsumer()`.

felixschlegel: Right, also applied this to `KafkaProducer.makeProducer`

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -166,6 +182,91 @@ public final class KafkaConsumer: Sendable, Service {
         }
     }
 
+    /// Initialize a new ``KafkaConsumer``.
+    ///
+    /// This factory method creates a consumer without that does not listen to any events other than consumer messages.
```
FranzBusch, line 187: Not a factory method any more

FranzBusch (approved):

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -166,6 +182,91 @@ public final class KafkaConsumer: Sendable, Service {
...
+        )
+
+        let consumer = try KafkaConsumer(
+            client: client,
+            stateMachine: stateMachine,
+            config: config,
+            logger: logger
+        )
+
+        let sourceAndSequence = NIOAsyncSequenceProducer.makeSequence(
```
blindspotbounty, line 260: I am not sure I fully understand an idea here: it seems like nothing is supposed to be supplied in this source...
Shouldn't be `sourceAndSequence.source` assigned anywhere?

felixschlegel: Yes, the primary intention of this PR was to expose the `KafkaConsumerEvents` itself without serving any events at the moment.

Following onto that PR events like rebalance etc. can be easily added into that existing `KafkaConsumerEvents` sequence.

Issue #78, opened 2023-06-29
Transactional API


mr-swifter:
Transactional API is required to use Exactly Once Semantics provided by Kafka.

One of the ideas how to structure this: 
- Make new actor `KafkaTransactionalProducer` which supports regular `KafkaProducer` API and transactional API (like `beginTransaction`, `commitTransaction`, `abortTransaction` etc. Maybe it's simpler to make `KafkaProducer` as class and inherit transactional producer from it to avoid code duplication
- - `KafkaTransactionalProducer` should have `send` and `sendOffset` within a transaction
- `KafkaTransactionalProducer` should take `transactional.id` as parameter.
- `KafkaTransactionalProducer` should call `rd_kafka_init_transactions(...)` and make sure it's initialised and not fenced with others.
- `KafkaTransactionalProducer` should handle retriable errors and tries to recover. Possible such errors also need to be delivered to optional callback (as notification)

blindspotbounty:
We had a couple of discussions with @mr-swifter regarding potential transactional API.

The easiest thing would be to implement calls "as is". However, there are some logic that might be used by more or less everyone such as:
1. Retry committing transaction if error is retriable `rd_kafka_error_is_retriable(error) == 1`
2. Abort transaction if transaction require abort and return corresponding error `if rd_kafka_error_txn_requires_abort(error) == 1`

Omitting insignificant  implementation details and making it more abstract, we were using the following code in send/commit and partially in abort transaction.
```swift
func commitTransaction(attempts: UInt, timeoutMs: Int = -1 /* wait until transaction.timeout.ms */) async -> Result<Void, KafkaError> {
        for _ in 0..<attempts {
            let error = await forBlockingFunc {
                rd_kafka_commit_transaction(self.kafkaHandle, Int32(timeoutMs))
            }

             /* check if transaction is completed successfully  */
             if error == nil { return .success(()) }

             /* destroy error in any case */
             defer { rd_kafka_error_destroy(error) } 

             /* check if transaction is retriable and retry */
             if rd_kafka_error_is_retriable(error) == 1 { continue }

             /* check if transaction need to be aborted */
             if rd_kafka_error_txn_requires_abort(error) == 1 {
                 // at this point we cannot retry, client code should begin transaction from scratch
                 let res = await abortTransaction()
                 return /* TransactionAbortedError/InconsistentStateError from `res` */
             }

             let description = String(cString: rd_kafka_error_string(error))
[...]

PR #98, opened 2023-07-24
add approx. transactional api

blindspotbounty:
There is an approximate solution for Transactional API for Kafka gsoc.

Some ideas that were expressed in https://github.com/swift-server/swift-kafka-gsoc/issues/78#issuecomment-1642042408

There are the following ideas:
1. We might initTransaction and switch state in machine
2. Run blocking transactional calls in GlobalQueue
3. Retry retryable errors
4. Automatically abort transactions when abort error received

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -44,6 +44,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.55.0"),
         .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.0.0-alpha.1"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.0.0"),
+        .package(url: "https://github.com/ordo-one/package-concurrency-helpers", .upToNextMajor(from: "1.0.0")),
         // The zstd Swift package produces warnings that we cannot resolve:
         // https://github.com/facebook/zstd/issues/3328
         .package(url: "https://github.com/facebook/zstd.git", from: "1.5.0"),
@@ -73,6 +74,7 @@ let package = Package(
             name: "SwiftKafka",
             dependencies: [
                 "Crdkafka",
+                .product(name: "ConcurrencyHelpers", package: "package-concurrency-helpers", moduleAliases: ["ConcurrencyHelpers" : "BlockingCallWrapper"]),
                 .product(name: "NIOCore", package: "swift-nio"),
                 .product(name: "ServiceLifecycle", package: "swift-service-lifecycle"),
                 .product(name: "Logging", package: "swift-log"),
--- a/Sources/SwiftKafka/Configuration/KafkaProducerConfiguration.swift
+++ b/Sources/SwiftKafka/Configuration/KafkaProducerConfiguration.swift
@@ -36,6 +36,8 @@ public struct KafkaProducerConfiguration {
     /// When set to true, the producer will ensure that messages are successfully produced exactly once and in the original produce order. The following configuration properties are adjusted automatically (if not modified by the user) when idempotence is enabled: max.in.flight.requests.per.connection=5 (must be less than or equal to 5), retries=INT32_MAX (must be greater than 0), acks=all, queuing.strategy=fifo. Producer instantation will fail if user-supplied configuration is incompatible.
     /// Default: `false`
     public var enableIdempotence: Bool = false
+    
+    public var transactionalId: String?
 
     /// Producer queue options.
     public var queue: KafkaConfiguration.QueueOptions = .init()
@@ -108,6 +110,9 @@ extension KafkaProducerConfiguration {
         var resultDict: [String: String] = [:]
 
         resultDict["enable.idempotence"] = String(self.enableIdempotence)
+        if let transactionalId {
+            resultDict["transactional.id"] = transactionalId
+        }
         resultDict["queue.buffering.max.messages"] = String(self.queue.bufferingMaxMessages)
         resultDict["queue.buffering.max.kbytes"] = String(self.queue.bufferingMaxKBytes)
         resultDict["queue.buffering.max.ms"] = String(self.queue.bufferingMaxMilliseconds)
--- a/Sources/SwiftKafka/KafkaError.swift
+++ b/Sources/SwiftKafka/KafkaError.swift
@@ -54,7 +54,7 @@ public struct KafkaError: Error, CustomStringConvertible {
     }
 
     static func rdKafkaError(
-        wrapping error: rd_kafka_resp_err_t, file: String = #fileID, line: UInt = #line
+        wrapping error: rd_kafka_resp_err_t, isFatal: Bool = false, file: String = #fileID, line: UInt = #line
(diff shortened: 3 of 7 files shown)


Sources/SwiftKafka/KafkaError.swift
```diff
@@ -162,6 +192,10 @@ extension KafkaError {
             case messageConsumption
             case topicCreation
             case topicDeletion
+            case transactionAborted
```
blindspotbounty, line 195: I was thinking if we should add subcodes or provide code from Kafka as is..

Package.swift
```diff
@@ -44,6 +44,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.55.0"),
         .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.0.0-alpha.1"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.0.0"),
+        .package(url: "https://github.com/ordo-one/package-concurrency-helpers", .upToNextMajor(from: "1.0.0")),
```
FranzBusch, line 47: We can't add this dependency.

Sources/SwiftKafka/Configuration/KafkaProducerConfiguration.swift
```diff
@@ -36,6 +36,8 @@ public struct KafkaProducerConfiguration {
     /// When set to true, the producer will ensure that messages are successfully produced exactly once and in the original produce order. The following configuration properties are adjusted automatically (if not modified by the user) when idempotence is enabled: max.in.flight.requests.per.connection=5 (must be less than or equal to 5), retries=INT32_MAX (must be greater than 0), acks=all, queuing.strategy=fifo. Producer instantation will fail if user-supplied configuration is incompatible.
     /// Default: `false`
     public var enableIdempotence: Bool = false
+    
+    public var transactionalId: String?
```
FranzBusch, line 40: We need some comments on this and what it does. If it only makes sense for the transaction producer we should not expose it all in this configuration and maybe have to introduce a new `KafkaTransactionalProducerConfiguration`

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -261,6 +261,42 @@ public final class KafkaProducer: Service, Sendable {
             return KafkaProducerMessageID(rawValue: newMessageID)
         }
     }
+    
+    func initTransactions(timeout: Duration = .seconds(5)) async throws {
```
FranzBusch, line 265: I think we definitely want to create a new `KafkaTransactionalProducer` type so that the APIs stay clear and separated between the normal producer and the transaction one.

Additionally, I think the APIs should look a bit different. I can imagine something like this

```
public func withTransaction<Result>(_ body: @Sendable (KafkaTransaction) async throws -> Void) async throws -> Result {
    // init transaction
    // begin transaction
    // call out to user code and pass a new `Transaction` struct to them
    // at the end call abort or finish transaction for them
}

```

Package.swift
blindspotbounty: Sure, I can copy helpers from there.

Sources/SwiftKafka/KafkaProducer.swift
blindspotbounty: That makes perfect sense to have a separate producer for transactions. 
We have it at a higher level but that might be done here

Sources/SwiftKafka/Configuration/KafkaProducerConfiguration.swift
blindspotbounty: Do you think of some wrapper around producer configuration like the following?
```
struct KafkaTransactionalProducerConfiguration {
    var producerConfiguration: KafkaTransactionalProducerConfiguration
    public var transactionalId: String
}
```

FranzBusch: I am not sure that works. Presumably there are some configurations that don't apply to a transactional producer. We probably should introduce a new configuration and just duplicate stuff like we did for the consumer and producer.

Sources/SwiftKafka/Configuration/KafkaProducerConfiguration.swift
```diff
@@ -158,6 +117,8 @@ extension KafkaProducerConfiguration: Hashable {}
 
 extension KafkaProducerConfiguration: Sendable {}
 
+extension KafkaProducerConfiguration: KafkaProducerSharedProperties {}
```
blindspotbounty, line 120: I've tried to use at least internal shared protocol to unite mostly all properties in producer/transactional producer instead of copying them

Sources/SwiftKafka/Utilities/BlockingCall.swift
```diff
@@ -0,0 +1,10 @@
+import Dispatch
+
+// performs blocking calls outside of cooperative thread pool
+internal func performBlockingCall<T>(queue: DispatchQueue, body: @escaping () -> T) async -> T {
+    await withCheckedContinuation { continuation in
```
blindspotbounty, line 5: I am not sure that there is a good way to use swift concurrency with blocking calls.
The only thing is to use some separate thread to avoid contract violation

Tests/IntegrationTests/SwiftKafkaTests.swift
```diff
@@ -489,4 +494,115 @@ final class SwiftKafkaTests: XCTestCase {
             XCTAssertTrue(acknowledgedMessages.contains(where: { $0.value == ByteBuffer(string: message.value) }))
         }
     }
+
+    func testProduceAndConsumeWithTransaction() async throws {
```
blindspotbounty, line 498: This is just an example test so far to demonstrate how current API would look like from user perpective

Sources/SwiftKafka/Configuration/KafkaTransactionalProducerConfiguration.swift
```diff
@@ -0,0 +1,134 @@
...
+                "Flush timeout outside of valid range \(0...Int32.max)"
+            )
+        }
+    }
+
+    // MARK: - Producer-specific Config Properties
+
+    /// When set to true, the producer will ensure that messages are successfully produced exactly once and in the original produce order. The following configuration properties are adjusted automatically (if not modified by the user) when idempotence is enabled: max.in.flight.requests.per.connection=5 (must be less than or equal to 5), retries=INT32_MAX (must be greater than 0), acks=all, queuing.strategy=fifo. Producer instantation will fail if user-supplied configuration is incompatible.
+    /// Default: `false`
+    internal let enableIdempotence: Bool = true
```
blindspotbounty, line 40: The difference between `KafkaProducerConfiguration` and `KafkaTransactionalProducerConfiguration` is in couple of properties:
1. `enableIdempotence` always `true` for transactions
2. `maxInFlightRequestsPerConnection` not greater than `5`
3. `transactionsTimeout` is set to socket timeout or greater than socket timeout

PR #104, opened 2023-07-28
Public API v1

felixschlegel:
### Modifications:

* updated public API for:
    * `KafkaProducer`
    * `KafkaProducerMessage`
    * `KafkaContiguousBytes`
    * `KafkaProducerMessageID`
    * `KafkaProducerConfiguration`
    * `KafkaDeliveryReport`
    * `KafkaAcknowledgedMessage`
    * `KafkaConsumer`
    * `KafkaConsumerMessage`
    * `KafkaConsumerConfiguration`
    * `KafkaTopicConfiguration`
    * `KafkaPartition`
    * `KafkaConfiguration`
    * `KafkaConfiguration+Security`

### DIFF
--- a/Sources/KafkaFoundationCompat/Data+KafkaContiguousBytes.swift
+++ b/Sources/KafkaFoundationCompat/Data+KafkaContiguousBytes.swift
@@ -12,15 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if swift(<5.9)
-@preconcurrency import Foundation
-#else
-import Foundation
-#endif
+import struct Foundation.Data
 import Kafka
-
-#if swift(<5.9)
-extension Data: @unchecked Sendable {}
-#endif
 
 extension Data: KafkaContiguousBytes {}
--- a/Sources/KafkaFoundationCompat/Data+KafkaContiguousBytes.swift
+++ b/Sources/KafkaFoundationCompat/Data+KafkaContiguousBytes.swift
@@ -12,7 +12,15 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if swift(<5.9)
+@preconcurrency import Foundation
+#else
 import Foundation
+#endif
 import Kafka
+
+#if swift(<5.9)
+extension Data: @unchecked Sendable {}
+#endif
 
 extension Data: KafkaContiguousBytes {}
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ The `send(_:)` method of `KafkaProducer` returns a message-id that can later be 
 ```swift
 let broker = KafkaConfiguration.Broker(host: "localhost", port: 9092)
 var config = KafkaProducerConfiguration()
-config.bootstrapServers = [broker]
+config.bootstrapBrokerAddresses = [broker]
 
 let (producer, events) = try KafkaProducer.makeProducerWithEvents(
     config: config,
@@ -86,7 +86,7 @@ var config = KafkaConsumerConfiguration(
         topic: "topic-name"
     )
 )
-config.bootstrapServers = [broker]
+config.bootstrapBrokerAddresses = [broker]
 
 let consumer = try KafkaConsumer(
     config: config,
@@ -123,7 +123,7 @@ let broker = KafkaConfiguration.Broker(host: "localhost", port: 9092)
 var config = KafkaConsumerConfiguration(
     consumptionStrategy: .group(id: "example-group", topics: ["topic-name"])
 )
-config.bootstrapServers = [broker]
+config.bootstrapBrokerAddresses = [broker]
 
 let consumer = try KafkaConsumer(
     config: config,
@@ -161,7 +161,7 @@ var config = KafkaConsumerConfiguration(
     consumptionStrategy: .group(id: "example-group", topics: ["topic-name"])
 )
 config.enableAutoCommit = false,
-config.bootstrapServers = [broker]
+config.bootstrapBrokerAddresses = [broker]
 
 let consumer = try KafkaConsumer(
     config: config,
--- a/Sources/Kafka/Configuration/KafkaConfiguration+Security.swift
+++ b/Sources/Kafka/Configuration/KafkaConfiguration+Security.swift
@@ -17,7 +17,7 @@ extension KafkaConfiguration {
 
     /// Use to configure an TLS connection.
     public struct TLSConfiguration: Sendable, Hashable {
-        /// Certificate chain consisting of one leaf certificate and potenentially multiple intermediate certificates.
+        /// Certificate chain consisting of one leaf certificate and potentially multiple intermediate certificates.
         /// The public key of the leaf certificate will be used for authentication.
         public struct LeafAndIntermediates: Sendable, Hashable {
             internal enum _Key: Sendable, Hashable {
@@ -27,7 +27,7 @@ extension KafkaConfiguration {
 
             let _internal: _Key
 
-            /// Read certificate chain from file.
+            /// Read certificate chain from a file.
             public static func file(location: String) -> LeafAndIntermediates {
                 return LeafAndIntermediates(
                     _internal: .file(location: location)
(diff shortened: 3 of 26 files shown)


Sources/SwiftKafka/Utilities/Duration+Helpers.swift
```diff
@@ -0,0 +1,25 @@
...
+
+extension Duration {
+    internal var inMilliseconds: UInt {
+        let seconds = Double(components.seconds) * 1000.0
+        let attoseconds = Double(components.attoseconds) * 1e-15
+        return UInt(seconds + attoseconds)
+    }
+
+    internal var canBeRepresentedAsMilliseconds: Bool {
+        return self.inMilliseconds > 0
```
blindspotbounty, line 23: Not sure if it is right place but in some places zero is a valid milliseconds as well
should be `self == .zero || self.inMilliseconds > 0`?

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -128,7 +128,7 @@ public final class KafkaConsumer: Sendable, Service {
     /// State of the `KafkaConsumer`.
     private let stateMachine: NIOLockedValueBox<StateMachine>
 
-    /// `AsyncSequence` that returns all ``KafkaConsumerMessage`` objects that the consumer receives.
+    /// An asynchronous sequence containing messages from the Kafka cluster.
     public let messages: KafkaConsumerMessages
```
blindspotbounty, line 132: Shouldn't it be `KafkaConsumerEvents` including messages as a part of those events?

Sources/SwiftKafka/KafkaConsumer.swift
```diff
@@ -295,20 +295,20 @@ public final class KafkaConsumer: Sendable, Service {
     private func assign(
```
blindspotbounty, line 295: I believe that `assign` and especially `subscribe` should be public.
There could be a valid scenario when topics are not known (or not all topics/partitions known) at the point of consumer creation. If method is private, I see two possible solutions:
1. Re-create whole `KafkaConsumer` with newly known topics -> slow
2. Create additional `KafkaConsumer` for new topics -> thread explosion by librdkafka 

If allow to use `subscribe/unsubscribe`, it would be possible to share one consumer across multiple topics/partitions. Otherwise, if topics are unknown or partially unknown at startup, it would require to always create consumer per new topic or re-create it.

Sources/SwiftKafka/KafkaProducer.swift
```diff
@@ -238,12 +235,15 @@ public final class KafkaProducer: Service, Sendable {
...
     ///
-    /// - Parameter message: The ``KafkaProducerMessage`` that is sent to the KafkaCluster.
-    /// - Returns: Unique ``KafkaProducerMessageID``matching the ``KafkaAcknowledgedMessage/id`` property
-    /// of the corresponding ``KafkaAcknowledgedMessage``.
+    /// - Parameter message: The ``KafkaProducerMessage`` to send.
+    /// - Returns: Unique ``KafkaProducerMessageID``matching the ``KafkaDeliveryReport/id`` property
+    /// of the corresponding ``KafkaDeliveryReport``.
     /// - Throws: A ``KafkaError`` if sending the message failed.
     @discardableResult
     public func send<Key, Value>(_ message: KafkaProducerMessage<Key, Value>) throws -> KafkaProducerMessageID {
```
blindspotbounty, line 249: I was thinking that we have two places to receive error: 
1. Here `send` method can throw an exception
2. And `.failure` can be received in `events`

Would be nice to unite these things

Sources/SwiftKafka/Configuration/KafkaConfiguration.swift
```diff
@@ -14,159 +14,231 @@
...
 
         public var description: String {
             "\(self.host):\(self.port)"
         }
 
-        public init(host: String, port: Int) {
-            self.host = host
-            self.port = port
-        }
+        public init() {}
```
FranzBusch, line 29: I think for this particular struct we should keep the init and not add any defaults since they don't make much sense or are these the defaults from rdkafka itself?

Sources/SwiftKafka/Data/KafkaContiguousBytes.swift
```diff
@@ -13,14 +13,14 @@
...
-/// to be used for serialization of Kafka messages.
+/// to be used for the serialization of Kafka messages.
 /// It provides a general interface for bytes since the Swift Standard Library currently does not
 /// provide such a protocol.
 ///
 /// By conforming your own types to this protocol, you will be able to pass instances of said types
 /// directly to ``KafkaProducerMessage`` as key and value.
-public protocol KafkaContiguousBytes {
-    /// Calls the given closure with the contents of underlying storage.
+public protocol KafkaContiguousBytes: Sendable, Hashable {
```
FranzBusch, line 22: Why do we need those two inheritances here?

Sources/SwiftKafka/KafkaConsumer.swift
FranzBusch: For now we keep it separate because we expect the the messages to be the default case and rarely people want to use the other events.

Sources/SwiftKafka/KafkaConsumer.swift
FranzBusch: I understand the need for this but I am not sure if we should provide those and if we can do so safely. The problem is that if we allow manual change of topics/partitions then we might still have messages buffered from the previous and it might be confusing for the user.

At least we should track this in a separate issue and not make it part of this PR.

Sources/SwiftKafka/KafkaProducer.swift
FranzBusch: I don't think we can or should unite those two. The failure on `send` here is when we fail to enqueue a message. The failure on the events is if we failed to deliver/acknowledge the message.

Sources/SwiftKafka/Utilities/Duration+Helpers.swift
FranzBusch: I would keep it like this for now and if we come across a concrete use-case where we need 0 then we should evaluate this again.

Sources/SwiftKafka/Data/KafkaContiguousBytes.swift
felixschlegel: Because `KafkaProducerMessage` is also `Sendable` and `Hashable`. However we can just get rid of that conformance if you like to

FranzBusch: Fair enough. Let's keep it!

FranzBusch (approved):

FranzBusch:
@felixschlegel pls fix the failing CI

Sources/KafkaFoundationCompat/Data+KafkaContiguousBytes.swift
```diff
@@ -12,7 +12,15 @@
...
+#if swift(<5.9)
+@preconcurrency import Foundation
+#else
 import Foundation
+#endif
 import Kafka
 
+#if swift(<5.9)
+extension Data: @unchecked Sendable {}
+#endif
```
FranzBusch, line 24: Is this really required. I was hoping a `@preconcurrency import struct Foundation.Data` would be enough. I don't think we can just conform `Data` to be `Sendable` here.

felixschlegel: Yes this is the error I got when only doing the `@preconcurrency import`:

https://ci.swiftserver.group/job/swift-kafka-client-swift58-prb/208/console

FranzBusch: That errors stems from you trying to conform `Data` to `Sendable`. My assumption is we don't have to do that at all just import with pre concurrency should be enough

FranzBusch (approved):

Sources/SwiftKafka/KafkaConsumer.swift
blindspotbounty: There are some thought on that.
I believe that your comment is more related to unassign and unsubscribe. However, `assign/incremental_assign/incremental_unassign` are used for rebalance. Which in its turn used pretty widely. Therefore, if it is planned, it might be nice to keep it in mind (probably it be relates to other comment for two streams for events/messages if sync is important to do from the library perspective). 

Anyway I was more concerned regarding `subscribe` method that might be helpful and should not contradict with safety.

felixschlegel: Yes definitely! Feel free to create an issue for that so we can kick off a more detailed discussion there!

FranzBusch: After giving this more thought we 100% cannot expose a subscribe and assign with the current APIs. This would mean users would expect to iterate the `messages` async sequence more than once (a new iteration after each subscribe/assign) call which is not allowed right now.

FranzBusch: What I would like to see is a concrete use-case where this is required which is not motivated by the performance of creating new consumers but rather from business logic needs.

blindspotbounty: I am pretty sure that all business logic needs can be covered by creating/removing consumers.

PR #107, opened 2023-07-31
avoid setting default values to topic configuration

blindspotbounty:
This is the fix for https://github.com/swift-server/swift-kafka-client/issues/106

Effectively, we should check that values are default and don't modify them. This way librdkafka would use default values from the main configuration.


Sources/SwiftKafka/RDKafka/RDKafkaTopicConfig.swift
```diff
@@ -35,6 +35,27 @@ struct RDKafkaTopicConfig {
...
+                    UnsafeMutableRawBufferPointer(
+                        start: configValue,
+                        count: sizeNoNullTerm))
+                _ = $0.initialize(from: buf)
+               return sizeNoNullTerm
+            }
+            if wasVal == value {
+                return // Values are equal, avoid changing (don't mark config as modified)
+            }
+        }
```
FranzBusch, line 57: @felixschlegel Can you take a look at this. I am not super happy with going through C here to figure this out.

Do we only have to do this for the topic configuration? If so can we just do it on the Swift side and avoid setting anything that is the default value?

PR #108, opened 2023-08-07
Rename to `swift-kafka-client`

felixschlegel:
### Modifications:

* change project name to `swift-kafka-client`
* replace package name `SwiftKafka` with `Kafka`


Tests/IntegrationTests/KafkaTests.swift
```diff
@@ -1,21 +1,21 @@
...
 // See LICENSE.txt for license information
-// See CONTRIBUTORS.txt for the list of swift-kafka-gsoc project authors
+// See CONTRIBUTORS.txt for the list of swift-kafka-client project authors
 //
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
 
 import struct Foundation.UUID
+@testable import Kafka
```
FranzBusch, line 16: Can we check in a follow up if we can get rid of all the `@testable` imports and just test the public API.

FranzBusch (approved):

Issue #110, opened 2023-08-09
Kafka consumer leaves group after 5 mins


blindspotbounty:
KafkaConsumer stops after 5 mins of consumption with error:
```
MAXPOLL] [thrd:main]: Application maximum poll interval (300000ms) exceeded by 472ms (adjust max.poll.interval.ms for long-running message processing): leaving group
```
There are two problems:
1. The error itself
2. This error is not reported to client code (i.e. not handled in `eventPoll`)

## How to reproduce
As minimal reproducing snippet I can suggest the following diff to existing test to make sequence big enough
```
diff --git a/Sources/Kafka/RDKafka/RDKafkaClient.swift b/Sources/Kafka/RDKafka/RDKafkaClient.swift
index d18cd72..f0ceb94 100644
--- a/Sources/Kafka/RDKafka/RDKafkaClient.swift
+++ b/Sources/Kafka/RDKafka/RDKafkaClient.swift
@@ -183,6 +183,10 @@ final class RDKafkaClient: Sendable {
                 self.handleLogEvent(event)
             case .offsetCommit:
                 self.handleOffsetCommitEvent(event)
+            case .error:
+                let err = rd_kafka_event_error_string(event)
+                let error = String(cString: err!)
+                fatalError("Got an error for: \(error)")
             case .none:
                 // Finished reading events, return early
                 return events
diff --git a/Tests/IntegrationTests/KafkaTests.swift b/Tests/IntegrationTests/KafkaTests.swift
index b5c551e..a5edd5d 100644
--- a/Tests/IntegrationTests/KafkaTests.swift
+++ b/Tests/IntegrationTests/KafkaTests.swift
@@ -291,7 +291,7 @@ final class KafkaTests: XCTestCase {
     }
 
     func testCommittedOffsetsAreCorrect() async throws {
-        let testMessages = Self.createTestMessages(topic: self.uniqueTestTopic, count: 10)
+        let testMessages = Self.createTestMessages(topic: self.uniqueTestTopic, count: 10000000) // some endless sequence of messages
         let firstConsumerOffset = testMessages.count / 2
         let (producer, acks) = try KafkaProducer.makeProducerWithEvents(configuration: self.producerConfig, logger: .kafkaTest)
 
@@ -307,6 +307,9 @@ final class KafkaTests: XCTestCase {
             ),
             bootstrapBrokerAddresses: [self.bootstrapBrokerAddress]
         )
[...]

PR #113, opened 2023-08-09
Fix: Kafka consumer leaves group after 5 mins

felixschlegel:
### Motivation:

This PR fixes issue #110.

`KakfaConsumer`: by polling the `rd_kafka_queue_get_main` queue instead of the `rd_kafka_queue_get_consumer` queue,
the timer for `max.poll.interval.ms` did not get reset which eventually
resulted in a timeout despite polling. (See [`librdkafka`
documentation](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#acacdb55ae7cb6abfbde89621e512b078))

### Modifications:

* `RDKafkaClient`:
    * rename `mainQueue` to `queue`
    * use `rd_kafka_queue_get_consumer` instead of
      `rd_kafka_queue_get_main` for `KakfaConsumer` clients
      -> this will reset the timer for `max.poll.interval.ms` so that
         the consumer does not time out despite polling
    * invoke `rd_kafka_queue_destroy(self.queue)` on
      `RDKafkaClient.deinit` to loose reference to queue

### DIFF
--- a/Sources/Kafka/RDKafka/RDKafkaClient.swift
+++ b/Sources/Kafka/RDKafka/RDKafkaClient.swift
@@ -51,8 +51,11 @@ final class RDKafkaClient: Sendable {
                 // Polling the queue counts as a consumer poll, and will reset the timer for `max.poll.interval.ms`.
                 self.queue = consumerQueue
             } else {
-                // Getting consumer queue failed, use main queue.
-                self.queue = rd_kafka_queue_get_main(self.kafkaHandle)
+                fatalError("""
+                Internal error: failed to get consumer queue. \
+                A group.id should be set even when the client is not part of a consumer group. \
+                See https://github.com/edenhill/librdkafka/issues/3261 for more information.
+                """)
             }
         } else {
             self.queue = rd_kafka_queue_get_main(self.kafkaHandle)
@@ -446,8 +449,7 @@ final class RDKafkaClient: Sendable {
     ///
     /// Make sure to run poll loop until ``RDKafkaClient/consumerIsClosed`` returns `true`.
     func consumerClose() throws {
-        let consumerQueue = rd_kafka_queue_get_consumer(self.kafkaHandle)
-        let result = rd_kafka_consumer_close_queue(self.kafkaHandle, consumerQueue)
+        let result = rd_kafka_consumer_close_queue(self.kafkaHandle, self.queue)
         let kafkaError = rd_kafka_error_code(result)
         if kafkaError != RD_KAFKA_RESP_ERR_NO_ERROR {
             throw KafkaError.rdKafkaError(wrapping: kafkaError)
--- a/Sources/Kafka/RDKafka/RDKafkaClient.swift
+++ b/Sources/Kafka/RDKafka/RDKafkaClient.swift
@@ -19,37 +19,51 @@ import Logging
 /// Base class for ``KafkaProducer`` and ``KafkaConsumer``,
 /// which is used to handle the connection to the Kafka ecosystem.
 final class RDKafkaClient: Sendable {
+    // Default size for Strings returned from C API
+    static let stringSize = 1024
+
     /// Determines if client is a producer or a consumer.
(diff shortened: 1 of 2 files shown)


Sources/Kafka/RDKafka/RDKafkaClient.swift
```diff
@@ -19,37 +19,51 @@ import Logging
...
 
-        self.mainQueue = rd_kafka_queue_get_main(self.kafkaHandle)
+        if type == .consumer {
+            if let consumerQueue = rd_kafka_queue_get_consumer(self.kafkaHandle) {
+                // (Important)
+                // Polling the queue counts as a consumer poll, and will reset the timer for `max.poll.interval.ms`.
+                self.queue = consumerQueue
+            } else {
+                // Getting consumer queue failed, use main queue.
+                self.queue = rd_kafka_queue_get_main(self.kafkaHandle)
```
blindspotbounty, line 55: should we throw an exception in that case?

felixschlegel: I changed it to a `fatalError` now. Reaching this statement clearly indicates that no `group.id` has been set (which we should do in either case, see link in code) which indicates a programming error.

Thanks for pointing that out!

Sources/Kafka/RDKafka/RDKafkaClient.swift
```diff
@@ -19,37 +19,51 @@ import Logging
...
         logger: Logger
     ) {
         self.kafkaHandle = kafkaHandle
         self.logger = logger
 
-        self.mainQueue = rd_kafka_queue_get_main(self.kafkaHandle)
+        if type == .consumer {
+            if let consumerQueue = rd_kafka_queue_get_consumer(self.kafkaHandle) {
+                // (Important)
+                // Polling the queue counts as a consumer poll, and will reset the timer for `max.poll.interval.ms`.
```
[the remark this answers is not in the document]
felixschlegel: We currently invoke [`rd_kafka_poll_set_consumer`](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#a9bfa0a1dd3f866cbf0c82fc089bd7904) which is basically a wrapping a queue forward 😄 

I updated the comment so it becomes clearer.

I believe the documentation refers to the difference between `rd_kafka_poll` and `rd_kafka_consumer_poll`. However, as we are using the event-based [`rd_kafka_queue_poll`](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#a2f147ed1c554c9048893fb1adde86dfa) we should be fine here as this function can be used on any queue.

Sources/Kafka/RDKafka/RDKafkaClient.swift
```diff
@@ -19,37 +19,54 @@ import Logging
...
+                self.queue = consumerQueue
+            } else {
+                fatalError("""
+                Internal error: failed to get consumer queue. \
+                A group.id should be set even when the client is not part of a consumer group. \
+                See https://github.com/edenhill/librdkafka/issues/3261 for more information.
+                """)
```
FranzBusch, line 58: Are we making sure this group id is set?

felixschlegel: So for our two consumption strategies:

```swift
public static func partition(
    _ partition: KafkaPartition,
    topic: String,
    offset: KafkaOffset = .end
)
        
public static func group(id groupID: String, topics: [String]) 
```

We assign the `group.id` ourselves for `.partition` and for `.group` the user supplies us with a `String`. However, when this 'groupID` `String` is `""` we run into the `fatalError`

FranzBusch (approved):

Sources/Kafka/RDKafka/RDKafkaClient.swift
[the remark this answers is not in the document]
blindspotbounty: Sounds good!

PR #115, opened 2023-08-11
Enable Sendability checks on CI

felixschlegel:
> **Note**: merge #114 first to get the tests passing

### DIFF
--- a/docker/docker-compose.2204.57.yaml
+++ b/docker/docker-compose.2204.57.yaml
@@ -16,6 +16,7 @@ services:
     image: swift-kafka-client:22.04-5.7
     environment:
       - WARN_AS_ERROR_ARG=-Xswiftc -warnings-as-errors
+      - STRICT_CONCURRENCY_ARG=-Xswiftc -strict-concurrency=complete
       # - SANITIZER_ARG=--sanitize=thread # TSan broken still
 
   shell:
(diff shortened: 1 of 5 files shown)


FranzBusch:
I don't think this is right. We have to pass -enable-experimental-feature as well

felixschlegel:
> I don't think this is right. We have to pass -enable-experimental-feature as well

Do you think we have to?

I got the [previous builds](https://ci.swiftserver.group/job/swift-kafka-client-swift57-prb/226/) deliberately failing on CI before I fixed the sendability issues. Also, [swift-open-api-generator](https://github.com/apple/swift-openapi-generator) does not use `-enable-experimental-feature` either for their sendabiliy CI.

FranzBusch:
> > I don't think this is right. We have to pass -enable-experimental-feature as well
> 
> 
> 
> Do you think we have to?
> 
> 
> 
> I got the [previous builds](https://ci.swiftserver.group/job/swift-kafka-client-swift57-prb/226/) deliberately failing on CI before I fixed the sendability issues. Also, [swift-open-api-generator](https://github.com/apple/swift-openapi-generator) does not use `-enable-experimental-feature` either for their sendabiliy CI.


Yeah I might be mistaken and this actually works.

FranzBusch (approved):

PR #120, opened 2023-08-22
Kafka Message Header API

felixschlegel:
### Motivation:

Be able to attach headers to a `KafkaProducerMessage` and read out
headers attached to a `KafkaConsumerMessage`.

### Modifications:

* create new type `struct KafkaHeader` representing a key-value pair of
  `String` key and `ByteBuffer` value
* add property `headers: [KafkaHeader]` to `KafkaConsumerMessage` and
  `KafkaConsumerMessage`
* use `rd_kafka_produceva` (varidadic arguments) to produce messages as
  `rd_kafka_produce` did not support setting message headers
* create helper class `RDKafkaUnsafeProducerMessage` that helps
  configuring the varidadic argument array for `rd_kafka_produceva`
* add new test asserting that both producing and consuming messages with
  message headers works


Sources/Kafka/KafkaHeader.swift
```diff
@@ -0,0 +1,46 @@
...
+import NIOCore
+
+/// A structure representing a header for a Kafka message.
+/// Headers are key-value pairs that can be attached to Kafka messages to provide additional metadata.
+public struct KafkaHeader: Sendable, Hashable {
+    /// The key associated with the header.
+    public var key: String
+
+    /// The value associated with the header.
+    public var value: ByteBuffer? // TODO: KafkaContiguousBytes
```
felixschlegel, line 24: Ideally, we would do the same `KafkaContiguousBytes` dance here as in `KafkaProducerMessage`.
However, the problem is that `KafkaHeader` is used in an array, which means the entire array would have to be generic over one specific `KafkaContiguousBytes` type which feels wrong too.

Sources/Kafka/RDKafka/RDKafkaUnsafeProducerMessage.swift
```diff
@@ -0,0 +1,112 @@
...
+
+        self._internal[index].vtype = RD_KAFKA_VTYPE_OPAQUE
+        self._internal[index].u.ptr = opaque
+        index += 1
+
+        for header in headers {
+            self._internal[index].vtype = RD_KAFKA_VTYPE_HEADER
+
+            // Copy name C String (strdup) as we do not own the headers: [KafkaHeader] array
+            guard let headerNameCPointer = header.key.withCString(strdup) else {
```
felixschlegel, line 76: I am not really happy with doing the `strdup` here. However, we need to somehow need to pass a `char *` pointer which we get from the `headers: [KafkaHeader]` array (hence accessing the pointer inside of the closure is weird).

I had a couple of ideas to reduce copying here:

* write C glue code that takes the `headers: [KafkaHeader]` array and does essentially what `RDKakfaUnsafeProducerMessage` does, just in a more enclosed way
* use copy-on-write / a String pool to avoid unnecessary copies of the String pointer (if a message header name occurs more than once we can avoid the additional copy with our String pool)

Sources/Kafka/RDKafka/RDKafkaUnsafeProducerMessage.swift
```diff
@@ -0,0 +1,112 @@
...
+                continue
+            }
+            self._internal[index].u.header.name = UnsafePointer(headerNameCPointer)
+
+            if let value = header.value {
+                // Copy value bytes as we do not own the headers: [KafkaHeader] array
+                let valueCopyBuffer = UnsafeMutableRawPointer.allocate(byteCount: value.readableBytes, alignment: 1)
+                value.withUnsafeReadableBytes { valueBuffer in
+                    if let baseAddress = valueBuffer.baseAddress {
+                        valueCopyBuffer.copyMemory(from: baseAddress, byteCount: valueBuffer.count)
```
felixschlegel, line 86: Also want to avoid copying here, see comment about `strdup`

Sources/Kafka/KafkaHeader.swift
FranzBusch: I am fine to leave this with `ByteBuffer` for now. Restricting the Array to single type isn't great.

Sources/Kafka/RDKafka/RDKafkaClient.swift
```diff
@@ -124,47 +124,193 @@ final class RDKafkaClient: Sendable {
...
+        cHeaders: inout [(key: UnsafePointer<CChar>, value: UnsafeRawBufferPointer?)],
+        _ body: ([(key: UnsafePointer<CChar>, value: UnsafeRawBufferPointer?)]) throws -> T
+    ) rethrows -> T {
+        if kafkaHeaders.isEmpty {
+            // Base case: we have read all kafkaHeaders and now invoke the accessor closure
+            // that can safely access the pointers in cHeaders
+            return try body(cHeaders)
+        } else {
+            guard let kafkaHeader = kafkaHeaders.popLast() else {
+                fatalError("kafkaHeaders should not be nil")
```
FranzBusch, line 283: Can we swap this around so that we only guard. Then we don't need the `fatalError` at all

FranzBusch (approved):

PR #122, opened 2023-08-23
`KafkaConsumer`: change locality of message error

felixschlegel:
### Motivation:

If the `KafkaConsumer` receives an errorneous message, it should not
throw in the `func run()` method but rather in the
`KafkaConsumer.messages` `ThrowingSequence`.

### Modifications:

* make `KafkaConsumerMessages` `AsyncSequence` wrap a sequence that
  yields a `Result<>` type -> `KafkaConsumerMessages` unwraps this
  result type and either yields or throws


FranzBusch (approved):

FranzBusch:
@felixschlegel Please update

PR #126, opened 2023-08-29
`KafkaConsumer.commitAsync`

felixschlegel:
### Motivation:

Having `KafkaConsumer.commitSync` be `async` is not always convenient as
it suspends the `KafkaConsumer.messages` read loop and can therefore
lower throughput.

This PR introduces a new method `KafkaConsumer.commitAsync` that allows
users who don't care about the result of the `commit` to commit in a
"fire-and-forget" manner.

### Modifications:

* new method `KafkaConsumer.commitAsync`
* rename `KafkaConsumer.StateMachine.commitSync` to
  `KafkaConsumer.StateMachine.commit` to serve both `commitSync` and
  `commitAsync`
* add new test for `KafkaConsumer.commitAsync`

### DIFF
--- a/Sources/Kafka/KafkaConsumer.swift
+++ b/Sources/Kafka/KafkaConsumer.swift
@@ -359,6 +359,32 @@ public final class KafkaConsumer: Sendable, Service {
     }
 
     /// Mark all messages up to the passed message in the topic as read.
+    /// Schedules a commit and returns immediately.
+    /// Any errors encountered after scheduling the commit will be discarded.
+    ///
+    /// This method is only used for manual offset management.
+    ///
+    /// - Warning: This method fails if the ``KafkaConsumerConfiguration/isAutoCommitEnabled`` configuration property is set to `true` (default).
+    ///
+    /// - Parameters:
+    ///     - message: Last received message that shall be marked as read.
+    /// - Throws: A ``KafkaError`` if committing failed.
+    public func commitAsync(_ message: KafkaConsumerMessage) throws {
+        let action = self.stateMachine.withLockedValue { $0.commit() }
+        switch action {
+        case .throwClosedError:
+            throw KafkaError.connectionClosed(reason: "Tried to commit message offset on a closed consumer")
+        case .commit(let client):
+            guard self.configuration.isAutoCommitEnabled == false else {
+                throw KafkaError.config(reason: "Committing manually only works if isAutoCommitEnabled set to false")
+            }
+
+            try client.commitAsync(message)
+        }
+    }
+
+    /// Mark all messages up to the passed message in the topic as read.
+    /// Awaits until the commit succeeds or an error is encountered.
     ///
     /// This method is only used for manual offset management.
     ///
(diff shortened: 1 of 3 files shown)


blindspotbounty:
That make sense to have async method, however it is especially a problem with fixed poll interval. If commits are often and poll interval is adaptive, that should not be a problem (unless it is a rare operation).
We bumped into that problem in particular (https://github.com/swift-server/swift-kafka-client/issues/128) and made polls adaptive, that helps a lot and in most cases not require async commit.

FranzBusch:
@felixschlegel Can you elaborate the use-case for the async commit here a bit more? When would you want to use this while doing manual commit management?

felixschlegel:
> @felixschlegel Can you elaborate the use-case for the async commit here a bit more? When would you want to use this while doing manual commit management?

I don't know if this is the answer you're looking for, but generally manual commits allow the user to decide when a message is marked as processed.

The problem with `commitSync` is that it adds a lot of latency to the read loop as it waits for the commit to be complete before moving on.

With `commitAsync` you just trigger a commit without handling the result of the commit (although we could add this with a callback-based approach) to get higher throughput while still having the freedom to decide when a message is considered processed.

In the worst case, if `commitAsync` fails, either the offset of message `i` is committed by committing the offset of message `i + 1` or the offset of the message is not committed at all, which could lead to another consumer reading the message again.

FranzBusch (commented):
Thanks for explaining @felixschlegel. One last comment

What about naming these two methods `commit() async throws` and `triggerCommit() throws`?

Sources/Kafka/RDKafka/RDKafkaClient.swift
```diff
@@ -532,9 +532,40 @@ final class RDKafkaClient: Sendable {
...
     }
 
-    /// Non-blocking **awaitable** commit of a the `message`'s offset to Kafka.
+    /// Non-blocking "fire-and-forget" commit of a `message`'s offset to Kafka.
+    /// Schedules a commit and returns immediately.
+    /// Any errors encountered after scheduling the commit will be discarded.
     ///
     /// - Parameter message: Last received message that shall be marked as read.
+    /// - Throws: A ``KafkaError`` if scheduling the commit failed.
+    func commitAsync(_ message: KafkaConsumerMessage) throws {
```
FranzBusch, line 541: Can we align the name here with `triggerCommit` and `commit` please

FranzBusch (commented):
@felixschlegel Can you check why the CI is failing?

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -367,17 +369,46 @@ public final class KafkaConsumer: Sendable, Service {
     /// - Parameters:
     ///     - message: Last received message that shall be marked as read.
     /// - Throws: A ``KafkaError`` if committing failed.
+    public func triggerCommit(_ message: KafkaConsumerMessage) throws {
```
FranzBusch, line 372: Okay sorry for the bike shedding but we probably should call this `scheduleCommit` which has a bit better semantics. Can you rename the public and internal methods?

FranzBusch (approved):

PR #133, opened 2023-09-20
Consumer: `assign` / `subscribe` on `run()`

felixschlegel:
### Motivation:

We should start assignments / subscriptions when `KafkaConsumer.run()`
is invoked and not on initialization.

### Modifications:

* `KafkaConsumer`: move `assign` / `subscribe` from `init` to `func run()`

### DIFF
--- a/Sources/Kafka/KafkaConsumer.swift
+++ b/Sources/Kafka/KafkaConsumer.swift
@@ -181,13 +181,6 @@ public final class KafkaConsumer: Sendable, Service {
 
         // Forward main queue events to the consumer queue.
         try client.pollSetConsumer()
-
-        switch configuration.consumptionStrategy._internal {
-        case .partition(topic: let topic, partition: let partition, offset: let offset):
-            try self.assign(topic: topic, partition: partition, offset: offset)
-        case .group(groupID: _, topics: let topics):
-            try self.subscribe(topics: topics)
-        }
     }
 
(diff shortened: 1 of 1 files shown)


Sources/Kafka/KafkaConsumer.swift
```diff
@@ -331,6 +324,13 @@ public final class KafkaConsumer: Sendable, Service {
     }
 
     private func _run() async throws {
+        switch self.configuration.consumptionStrategy._internal {
+        case .partition(topic: let topic, partition: let partition, offset: let offset):
+            try self.assign(topic: topic, partition: partition, offset: offset)
+        case .group(groupID: _, topics: let topics):
+            try self.subscribe(topics: topics)
+        }
```
FranzBusch, line 332: We should probably go through the state machine here

felixschlegel: `func assign` and `func subscribe` already go through the state machine by invoking `stateMachine.setupConnection`

I could however hoist the state check here (so `state check -> check consumption strategy -> assign / subscribe` instead of `check consumption strategy -> state check -> assign / subscribe`

felixschlegel: Yes, it only works when in state `.initializing` 👍

FranzBusch (approved):

PR #135, opened 2023-10-03
Ship with BoringSSL

### LINKED ISSUE #123: rdkafka: SSL requirement
Currently, there is an implicit requirement to have openssl to build or use the library:
Possible build error:
```
In file included from /home/actions/_work/package-storage/package-storage/.build/checkouts/swift-kafka-client/Sources/Crdkafka/librdkafka/src/rdkafka_mock_handlers.c:34:
/home/actions/_work/package-storage/package-storage/.build/checkouts/swift-kafka-client/Sources/Crdkafka/librdkafka/src/rdkafka_int.h:55:10: fatal error: 'openssl/ssl.h' file not found
#include <openssl/ssl.h>
```
Possible runtime error for compiled code:
```
Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Termination Reason:    Namespace DYLD, Code 1 Library missing
```
Should OpenSSL library be statically linked or requirement explicitly mentioned in documentation?

felixschlegel:
Closes #123 

### Motivation:

No need to install `libressl` on machine to build project.

### Modifications:

* add dependency to `swift-nio-ssl` to obtain copy of `BoringSSL`
* create bridging headers allowing `librdkafka` to build with `BoringSSL` copy from `swift-nio-ssl`

#### Additional Changes:

* use `swift-service-lifecycle` version `2.1.0` instead of `beta`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -46,7 +46,8 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.55.0"),
-        .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.0.0-beta.1"),
+        .package(url: "https://github.com/apple/swift-nio-ssl", .upToNextMajor(from: "2.25.0")),
+        .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.1.0"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.0.0"),
         // The zstd Swift package produces warnings that we cannot resolve:
         // https://github.com/facebook/zstd/issues/3328
@@ -56,7 +57,7 @@ let package = Package(
         .target(
             name: "Crdkafka",
             dependencies: [
-                "COpenSSL",
+                .product(name: "NIOSSL", package: "swift-nio-ssl"),
                 .product(name: "libzstd", package: "zstd"),
             ],
             exclude: rdkafkaExclude,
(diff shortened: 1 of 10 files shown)


Package.swift
```diff
@@ -46,7 +46,8 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.55.0"),
-        .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.0.0-beta.1"),
+        .package(url: "https://github.com/apple/swift-nio-ssl", .upToNextMajor(from: "2.25.0")),
```
FranzBusch, line 49:
```suggestion
        .package(url: "https://github.com/apple/swift-nio-ssl", .from: "2.25.0"),
```

FranzBusch (commented):
Can you take a look at the CI failures?

felixschlegel:
> Can you take a look at the CI failures?

That's a weird one, I run into the exact issue as described here:

https://forums.swift.org/t/swift-5-9-compilation-failure-on-linux-when-using-system-modules-that-import-select-h/66275

Currently, my fix is to set the [`_GNU_SOURCE`](https://linux.die.net/man/7/feature_test_macros) macro when building the `Crdkafka` target, which gives me a lot of duplicate macro warnings but seems to work.

felixschlegel:
We need to get #137 merged before we get the `5.10` test to pass

FranzBusch (approved):

Issue #128, opened 2023-08-31
Make poll interval adaptive


blindspotbounty:
Currently, even if there are move than 100 events (which is hardcoded), we won't continue polling but call `Task.sleep()` which seems pretty strange.
I would like to suggest some adaptive polling interval, which would be depending on events: e.g.:
```
func eventPoll(events: inout [KafkaEvent], maxEvents: inout Int) -> Bool /* or enum { case shouldSleep, case shouldYield } */{
        events.removeAll(keepingCapacity: true)
        events.reserveCapacity(maxEvents)
        
        var shouldSleep = true

for ...
            switch eventType {
            case .deliveryReport:
                let forwardEvent = self.handleDeliveryReportEvent(event)
                events.append(forwardEvent)
                shouldSleep = false
            case .fetch:
                if let forwardEvent = self.handleFetchEvent(event) {
                    events.append(forwardEvent)
                    shouldSleep = false
                }
            case .log:
                self.handleLogEvent(event)
            case .offsetCommit:
                self.handleOffsetCommitEvent(event)
                shouldSleep = false
            case .statistics:
                events.append(self.handleStatistics(event))
            case .rebalance:
                events.append(self.handleRebalance(event))
                shouldSleep = false
            case .error:
                break
            case .none:
                // Finished reading events, return early
                return shouldSleep
            default:
                break // Ignored Event
            }
```
The idea is to then adjust polling time the following way:
```
            case .pollForAndYieldMessage(let client, let source, let eventSource):
                let shouldSleep = client.eventPoll(events: &events, maxEvents: &maxEvents)
                for event in events { ... }
                if shouldSleep {
                    pollInterval = min(self.configuration.pollInterval, pollInterval * 2)
                    try await Task.sleep(for: pollInterval)
                } else {
                    pollInterval = max(pollInterval / 3, .milliseconds(1))
                    await Task.yield()
                }
```

That might be also useful to adjust events size but not sure if it is feasible.

FranzBusch:
Could you elaborate why we might want to do this? Are you running into any problems with the current behaviour?

blindspotbounty:
Sure, this is primarily related to performance with `commitSync`.
Let's take some simple example:
```
for try await message in consumer.messages {
    print("Current ts: \(Date.now.timeIntervalSince1970)")
    try await consumer.commitSync(message)
}
```
All timestamps will differ with 100ms (default poll interval):
```
Current ts: 1694001547.020994
[...]

PR #139, opened 2023-10-11
`KafkaConsumer`: back pressure  + improved read speed

### LINKED ISSUE #131: Enable back pressure for KafkaConsumer
We performed some basic tests for consumer but with big number of records >11M
```
for try await record in consumer.messages {
    i = record.offset
    ctr += 1

    if ctr % 1000 == 0 {
        print("read up to \(i), ctr: \(ctr)")
    }
}
```
Unfortunately, without back pressure it leads to memory growth. For simplicity attaching screenshot:
memgrow

At some point, when all entries are read from kafka, memory usage is declining but would be nice if it would be possible to limit it with configuration (e.g. NIO HighLow watermark backpressure strategy).

PS: it seems that back pressure was removed around here https://github.com/swift-server/swift-kafka-client/pull/66

felixschlegel:
### Motivation:

Closes #131.

### Example:

I ran an example consumer reading a topic with approx. `11_000_000` messages:

```swift
var consumerConfig = KafkaConsumerConfiguration(
    consumptionStrategy: .partition(
        KafkaPartition(rawValue: 0),
        topic: "test-topic",
        offset: KafkaOffset(rawValue: 0) // Important: Read from beginning!
    ),
    bootstrapBrokerAddresses: [self.bootstrapBrokerAddress]
)
consumerConfig.pollInterval = .zero
consumerConfig.autoOffsetReset = .beginning // Always read topics from beginning
consumerConfig.broker.addressFamily = .v4

let consumer = try KafkaConsumer(
    configuration: consumerConfig,
    logger: .kafkaTest
)

let serviceGroupConfiguration = ServiceGroupConfiguration(services: [consumer], logger: .kafkaTest)
let serviceGroup = ServiceGroup(configuration: serviceGroupConfiguration)

try await withThrowingTaskGroup(of: Void.self) { group in
    // Run Task
    group.addTask {
        try await serviceGroup.run()
    }

    // Consumer Task
    group.addTask {
        var count = 0
        for try await message in consumer.messages {
            _ = message // drop message
            count += 1
            try await Task.sleep(for: .milliseconds(1))
            if count % 1000 == 0 {
                print(count)
            }
        }
    }

    // Wait for Consumer Task to complete
    try await group.next()
    // Shutdown the serviceGroup
    await serviceGroup.triggerGracefulShutdown()
}

```

##### Before

(Without back pressure)

Screenshot 2023-10-11 at 13 23 53

##### After

(With back pressure)

Screenshot 2023-10-11 at 20 48 18

This result is tolerable since the `queued.max.messages.kbytes` configuration property defaults to prefetching at max `~65` MegaBytes of messages. Exposing `queued.max.messages.kbytes` will be done in a follow-up PR.

### Modifications:

* re-add `KafkaConsumerConfiguration.backPressureStrategy: BackPressureStrategy`, currently allowing users to add high-low-watermark backpressure to their `KafkaConsumer`s
* `KafkaConsumer`:
    * make `KafkaConsumerMessages` use `NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark` as backpressure strategy
    * remove `rd_kafka_poll_set_consumer` -> use two separate queues for consumer events and consumer messages so we can exert backpressure on the consumer message queue
    * remove idle polling mechanism where incoming messages were discarded when `KafkaConsumerMessages` was terminated -> we now have to independent queues
    * rename `.pollForAndYieldMessage` -> `.pollForEventsAndMessages`
    * refactor `State` and add `ConsumerMessagesSequenceState`
* `KafkaProducer`:
    * rename `.consumptionStopped` -> `.eventConsumptionFinished`
* `RDKafkaClient`:
    * bring back `consumerPoll()` * `eventPoll()`: only queue main queue for events since consumer messages are now handled on a different queue

### DIFF
--- a/Sources/Kafka/Configuration/KafkaConsumerConfiguration.swift
+++ b/Sources/Kafka/Configuration/KafkaConsumerConfiguration.swift
@@ -23,6 +23,37 @@ public struct KafkaConsumerConfiguration {
     /// Default: `.milliseconds(100)`
     public var pollInterval: Duration = .milliseconds(100)
 
+    /// A struct representing different back pressure strategies for consuming messages in ``KafkaConsumer``.
+    public struct BackPressureStrategy: Sendable, Hashable {
+        enum _BackPressureStrategy: Sendable, Hashable {
+            case watermark(low: Int, high: Int)
+        }
+
+        let _internal: _BackPressureStrategy
+
+        private init(backPressureStrategy: _BackPressureStrategy) {
+            self._internal = backPressureStrategy
+        }
+
+        /// A back pressure strategy based on high and low watermarks.
+        ///
+        /// The consumer maintains a buffer size between a low watermark and a high watermark
+        /// to control the flow of incoming messages.
+        ///
+        /// - Parameter low: The lower threshold for the buffer size (low watermark).
+        /// - Parameter high: The upper threshold for the buffer size (high watermark).
+        public static func watermark(low: Int, high: Int) -> BackPressureStrategy {
+            return .init(backPressureStrategy: .watermark(low: low, high: high))
+        }
+    }
+
+    /// The backpressure strategy to be used for message consumption.
+    /// See ``KafkaConsumerConfiguration/BackPressureStrategy-swift.struct`` for more information.
+    public var backPressureStrategy: BackPressureStrategy = .watermark(
+        low: 10,
+        high: 50
+    )
+
     /// A struct representing the different Kafka message consumption strategies.
     public struct ConsumptionStrategy: Sendable, Hashable {
         enum _ConsumptionStrategy: Sendable, Hashable {
@@ -63,7 +94,7 @@ public struct KafkaConsumerConfiguration {
     }
 
     /// The strategy used for consuming messages.
-    /// See ``KafkaConfiguration/ConsumptionStrategy`` for more information.
+    /// See ``KafkaConsumerConfiguration/ConsumptionStrategy-swift.struct-swift.struct`` for more information.
     public var consumptionStrategy: ConsumptionStrategy
 
     // MARK: - Consumer-specific Config Properties
--- a/Sources/Kafka/KafkaConsumer.swift
+++ b/Sources/Kafka/KafkaConsumer.swift
@@ -38,7 +38,7 @@ extension KafkaConsumerEventsDelegate: NIOAsyncSequenceProducerDelegate {
 
 /// `NIOAsyncSequenceProducerDelegate` for ``KafkaConsumerMessages``.
 internal struct KafkaConsumerMessagesDelegate: Sendable {
-    let stateMachine: NIOLockedValueBox<KafkaConsumer.ConsumerMessagesStateMachine>
+    let stateMachine: NIOLockedValueBox<KafkaConsumer.StateMachine>
 }
 
 extension KafkaConsumerMessagesDelegate: NIOAsyncSequenceProducerDelegate {
@@ -47,7 +47,7 @@ extension KafkaConsumerMessagesDelegate: NIOAsyncSequenceProducerDelegate {
     }
 
     func didTerminate() {
-        self.stateMachine.withLockedValue { $0.finish() }
+        self.stateMachine.withLockedValue { $0.finishMessageConsumption() }
     }
 }
 
@@ -152,8 +152,6 @@ public final class KafkaConsumer: Sendable, Service {
     private let logger: Logger
     /// State of the `KafkaConsumer`.
     private let stateMachine: NIOLockedValueBox<StateMachine>
-    // TODO: refactor
-    private let consumerMessagesStateMachine: NIOLockedValueBox<ConsumerMessagesStateMachine>
 
     /// An asynchronous sequence containing messages from the Kafka cluster.
     public let messages: KafkaConsumerMessages
@@ -166,20 +164,17 @@ public final class KafkaConsumer: Sendable, Service {
     /// - Parameters:
     ///     - client: Client used for handling the connection to the Kafka cluster.
     ///     - stateMachine: The state machine containing the state of the ``KafkaConsumer``.
-    ///      TODO
     ///     - configuration: The ``KafkaConsumerConfiguration`` for configuring the ``KafkaConsumer``.
     ///     - logger: A logger.
     /// - Throws: A ``KafkaError`` if the initialization failed.
     private init(
         client: RDKafkaClient,
         stateMachine: NIOLockedValueBox<StateMachine>,
-        consumerMessagesStateMachine: NIOLockedValueBox<ConsumerMessagesStateMachine>,
         configuration: KafkaConsumerConfiguration,
         logger: Logger
     ) throws {
         self.configuration = configuration
         self.stateMachine = stateMachine
-        self.consumerMessagesStateMachine = consumerMessagesStateMachine
         self.logger = logger
 
         let sourceAndSequence = NIOThrowingAsyncSequenceProducer.makeSequence(
@@ -193,7 +188,7 @@ public final class KafkaConsumer: Sendable, Service {
                     )
                 }
             }(),
-            delegate: KafkaConsumerMessagesDelegate(stateMachine: self.consumerMessagesStateMachine)
+            delegate: KafkaConsumerMessagesDelegate(stateMachine: self.stateMachine)
         )
 
         self.messages = KafkaConsumerMessages(
@@ -201,7 +196,7 @@ public final class KafkaConsumer: Sendable, Service {
             wrappedSequence: sourceAndSequence.sequence
         )
 
-        self.consumerMessagesStateMachine.withLockedValue {
+        self.stateMachine.withLockedValue {
             $0.initialize(
                 client: client,
                 source: sourceAndSequence.source
@@ -235,13 +230,11 @@ public final class KafkaConsumer: Sendable, Service {
             logger: logger
         )
 
-        let stateMachine = NIOLockedValueBox(StateMachine(client: client))
-        let consumerMessagesStateMachine = NIOLockedValueBox(ConsumerMessagesStateMachine())
+        let stateMachine = NIOLockedValueBox(StateMachine())
 
         try self.init(
             client: client,
             stateMachine: stateMachine,
-            consumerMessagesStateMachine: consumerMessagesStateMachine,
             configuration: configuration,
             logger: logger
         )
@@ -277,13 +270,11 @@ public final class KafkaConsumer: Sendable, Service {
             logger: logger
         )
 
-        let stateMachine = NIOLockedValueBox(StateMachine(client: client))
-        let consumerMessagesStateMachine = NIOLockedValueBox(ConsumerMessagesStateMachine())
+        let stateMachine = NIOLockedValueBox(StateMachine())
 
         let consumer = try KafkaConsumer(
             client: client,
             stateMachine: stateMachine,
-            consumerMessagesStateMachine: consumerMessagesStateMachine,
             configuration: configuration,
             logger: logger
         )
@@ -320,8 +311,6 @@ public final class KafkaConsumer: Sendable, Service {
             }
             try client.subscribe(topicPartitionList: subscription)
         }
-        // TODO: refactor
-        self.consumerMessagesStateMachine.withLockedValue { $0.setUpConnection() }
     }
 
     /// Assign the``KafkaConsumer`` to a specific `partition` of a `topic`.
@@ -342,8 +331,6 @@ public final class KafkaConsumer: Sendable, Service {
             assignment.setOffset(topic: topic, partition: partition, offset: Int64(offset.rawValue))
             try client.assign(topicPartitionList: assignment)
         }
-        // TODO: refactor
-        self.consumerMessagesStateMachine.withLockedValue { $0.setUpConnection() }
     }
 
     /// Start the ``KafkaConsumer``.
@@ -385,7 +372,7 @@ public final class KafkaConsumer: Sendable, Service {
             // Run loop for consumer messages
             group.addTask {
                 while !Task.isCancelled {
-                    let nextAction = self.consumerMessagesStateMachine.withLockedValue { $0.nextPollLoopAction() }
+                    let nextAction = self.stateMachine.withLockedValue { $0.nextConsumerPollLoopAction() }
                     switch nextAction {
                     case .pollForAndYieldMessage(let client, let source):
                         // Poll for new consumer message.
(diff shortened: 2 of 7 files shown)


Sources/Kafka/Configuration/KafkaConsumerConfiguration.swift
```diff
@@ -23,6 +23,37 @@ public struct KafkaConsumerConfiguration {
...
+        public static func watermark(low: Int, high: Int) -> BackPressureStrategy {
+            return .init(backPressureStrategy: .watermark(low: low, high: high))
+        }
+    }
+
+    /// The backpressure strategy to be used for message consumption.
+    /// See ``KafkaConsumerConfiguration/BackPressureStrategy-swift.struct`` for more information.
+    public var backPressureStrategy: BackPressureStrategy = .watermark(
+        low: 10,
+        high: 50
```
felixschlegel, line 54: Are you happy with these default values?

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -334,22 +355,32 @@ public final class KafkaConsumer: Sendable, Service {
...
-                        _ = source.yield(result)
-                    default:
-                        break // Ignore
+            case .pollForEventsAndMessage(let client, let source):
+                // Event poll to serve any events queued inside of `librdkafka`.
+                _ = client.eventPoll()
+                // Poll for new consumer message.
+                var result: Result<KafkaConsumerMessage, Error>?
+                do {
+                    if let message = try client.consumerPoll() {
```
blindspotbounty, line 364: It seems that previously we would poll for up to 100 messages before sleep and it looks like now it would be just one message.

Sources/Kafka/Configuration/KafkaConsumerConfiguration.swift
blindspotbounty: I think these values should work good for a wide range of cases

FranzBusch: Fine by me as well. Kafka is all around people tuning the settings to fit their messages.

Sources/Kafka/KafkaConsumer.swift
FranzBusch, line 364: We have to do more here than just calling the two separate polls. We probably want to serve the consumer poll in a child task and make it completely driven by the backpressure of the async sequence. However, we need to keep shutdown in mind so we need to inform the child task about this.

Saying all of that I am not yet convinced it is the right solution. The one question that I am asking myself right now is how do we get notified when the consumer queue has more messages again when we do the poll based and decouple the two queue polls. Is there a callback in rdkafka that we can set once we get `nil` from `consumerPoll` so that we can enqueue ourselves?

Sources/Kafka/RDKafka/RDKafkaClient.swift
```diff
@@ -451,18 +411,30 @@ final class RDKafkaClient: Sendable {
...
+    /// - Throws: A ``KafkaError`` if the received message is an error message or malformed.
+    func consumerPoll() throws -> KafkaConsumerMessage? {
+        guard let messagePointer = rd_kafka_consumer_poll(self.kafkaHandle, 0) else {
+            // No error, there might be no more messages
+            return nil
         }
+
+        defer {
+            // Destroy message otherwise poll() will block forever
+            rd_kafka_message_destroy(messagePointer)
```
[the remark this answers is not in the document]
felixschlegel: Yes, please create a separate issue for that 😄

Sources/Kafka/KafkaConsumer.swift
blindspotbounty: I believe there is no such notification in librdkafka. I guess the desired thing to do with native API is to use blocking call for consumer poll which contradicts with swift concurrency contract.
From the librdkafka sample (https://github.com/confluentinc/librdkafka/blob/master/examples/consumer.c#L209C17-L209C55):
```
for (;;) {
     const auto * msg = rd_kafka_consumer_poll(consumer.kafkaHandle, 100 /* 100ms timeout */);
     if msg == nullptr {
         continue
     }
     // ...
}
```
Other thought after confirming with librdkafka documentation is that all callbacks are called on queue's polls (https://github.com/confluentinc/librdkafka/blob/master/INTRODUCTION.md#threads-and-callbacks). That, unfortunately, makes any callbacks without polls useless.

I guess making dedicated thread for blocking polls is not inline with swift evolution. However, with swift concurrency probably it is okay to sleep/yield task for some configurable timeout (e.g. 10ms) if we have no more messages to read otherwise continue to read until bump into backpressure. Additionally, it is possible to make poll intervals adaptive depending on message flow (#128).

felixschlegel:
Hey folks,

I have implemented your requested changes:

* have two poll loops, one for normal events (every `pollInterval` time units) and one for consumer messages (read as long as you can without sleeping)

The poll loop for consumer messages follows a similar approach to what was proposed in #128 .

Also, I want to add a benchmark test evaluating that we achieve good memory usage (back pressure) and good read speeds with this implementation so it would be good if we can get #140 over the line 😄 

Please let me know what you think!

Best,
Felix

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -331,26 +352,89 @@ public final class KafkaConsumer: Sendable, Service {
...
+                try await self.eventRunLoop()
+            }
+
+            group.addTask {
+                try await self.messageRunLoop()
+            }
+
+            // Throw when one of the two child task throws
+            try await group.next()
+            try await group.next()
```
FranzBusch, line 366: We probably want to do a `group.cancelAll()` after the first returned without throwing.

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -331,26 +352,89 @@ public final class KafkaConsumer: Sendable, Service {
...
+                }
+                if let result {
+                    let yieldResult = source.yield(result)
+                    switch yieldResult {
+                    case .produceMore:
+                        break
+                    case .stopProducing:
+                        self.stateMachine.withLockedValue { $0.stopProducing() }
+                    case .dropped:
+                        break
```
FranzBusch, line 408: Here we probably want to `return` right?

Sources/Kafka/KafkaConsumer.swift
FranzBusch: If I read this correctly we are polling a single message and then yielding a single message right? It would probably be better if we read them in batches of 100 and then yield all of them at once. Would mean we have to acquire the locks a lot less.

Sources/Kafka/KafkaProducerEvent.swift
```diff
@@ -23,8 +23,6 @@ public enum KafkaProducerEvent: Sendable, Hashable {
         switch event {
         case .deliveryReport(results: let results):
```
blindspotbounty, line 24: Hm.. I guess it is out of scope here but should we have backpressure for delivery reports as well?

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -331,26 +352,89 @@ public final class KafkaConsumer: Sendable, Service {
...
+                    }
+
+                    self.stateMachine.withLockedValue { $0.newMessagesProduced() }
```
blindspotbounty, line 434: I guess It might make sense to continue without sleep after message was produced until we will out of messages again or bump into backpressure.

Sources/Kafka/KafkaConsumer.swift
blindspotbounty: Not sure it is entirely what I mean. This comment was related to changes before introducing task group.
There was a sleep after every consumed message.
So, it is not relevant now except probably this part: https://github.com/swift-server/swift-kafka-client/pull/139/files#r1360663871

Sources/Kafka/KafkaProducerEvent.swift
FranzBusch: I don't think we can since they happen on the `eventsQueue` unless there is a way to separate events to different queues.

felixschlegel: Yes, the problem is that we should call `poll()` at regular intervals to serve any queued callbacks/events. As @FranzBusch points out, we would need `librdkafka` to have a separate "delivery reports queue" for backpressure to make sense here since some events like log events should still be served even when our `events` `AsyncSequence` with all the `.deliveryReport` events is suspended.

blindspotbounty: Ah, yes, sure... We can only separate log queue from others...
So, we only can make some "unfair" backpressue, i.e. sometimes poll even if receive `stopProducing` which is dirty solution...

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -331,26 +352,89 @@ public final class KafkaConsumer: Sendable, Service {
...
+
+    /// Run loop polling Kafka for new events.
+    private func eventRunLoop() async throws {
+        while !Task.isCancelled {
+            let nextAction = self.stateMachine.withLockedValue { $0.nextEventPollLoopAction() }
+            switch nextAction {
+            case .pollForEvents(let client):
+                // Event poll to serve any events queued inside of `librdkafka`.
+                _ = client.eventPoll()
+                try await Task.sleep(for: self.configuration.pollInterval)
```
blindspotbounty, line 378: That is obviously out of scope regarding sleeping here and likely should be done separately. 
What do you think about symmetry with consumer polls thus sleeping only in case we out of events otherwise continue to poll events?

Sources/Kafka/KafkaProducerEvent.swift
felixschlegel: Yes, we could separate the log queue but then there is also other events like commit confirmation

blindspotbounty: Yeah, I got it, we may close this discussion, probably to think about it with separate case as it is out of scope for this PR anyway and might be not a problem (at least far).

Sources/Kafka/RDKafka/RDKafkaClient.swift
```diff
@@ -451,18 +411,30 @@ final class RDKafkaClient: Sendable {
...
+        }
+
+        // Reached the end of the topic+partition queue on the broker
+        if messagePointer.pointee.err == RD_KAFKA_RESP_ERR__PARTITION_EOF {
```
blindspotbounty, line 432: Hm, I wonder if now all errors will come here or some of them will be received in `eventsPoll()`

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -331,26 +352,89 @@ public final class KafkaConsumer: Sendable, Service {
...
+                    result = .failure(error)
+                }
+                if let result {
+                    self.stateMachine.withLockedValue { $0.produceMore() }
+                    let yieldResult = source.yield(result)
+                    switch yieldResult {
+                    case .produceMore:
+                        break
+                    case .stopProducing:
+                        self.stateMachine.withLockedValue { $0.stopProducing() }
```
blindspotbounty, line 429: It seems If we suspend producing here, we should not reach `newMessagesProduced`, otherwise we bump into `fatalError`:
https://github.com/swift-server/swift-kafka-client/pull/139/files/6686e982c3c1c905acc0a6f4653cdf903f6738ff#diff-8ba8c17c39d40fc29ec22c104d01bacc928df36e717000547f2a97d4203ef6a1R797

Sources/Kafka/KafkaConsumer.swift
felixschlegel: I'm not sure, what about the following case:

When the client application stops reading the `KafkaConsumerMessages` sequence the `messageRunLoop()` will return. However, the `eventRunLoop()` might still be processing a _consumer close_ so I don't see a benefit in cancelling here, but open to discussion

Sources/Kafka/KafkaConsumer.swift
felixschlegel: Yes, let's do this in a follow-up PR! This probably relates to the #128 issue

Sources/Kafka/KafkaConsumer.swift
felixschlegel: Yes indeed, very good catch!

felixschlegel:
We are failing on `5.10` because `OpaquePointer` is not `Sendable` — is that a deliberate change to Swift 5.10? The [documentation](https://developer.apple.com/documentation/swift/opaquepointer) still states that it should be `Sendable`

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -331,33 +352,112 @@ public final class KafkaConsumer: Sendable, Service {
...
     }
 
+    /// Read `maxMessages` consumer messages from Kafka.
+    ///
+    /// - Parameters:
+    ///     - client: Client used for handling the connection to the Kafka cluster.
+    ///     - maxMessages: Maximum amount of consumer messages to read in this invocation.
+    private func batchConsumerPoll(
+        client: RDKafkaClient,
+        maxMessages: Int = 100
```
blindspotbounty, line 438: We probably should provide this argument from poll based on backpressure. Currently, we have 50 messages high watermark as default value for backpressure, so we may break backpressure as twice if read 100 messages and enqueue them to stream. 
Not sure what is right here but maybe we can call it with one of those high or low watermark limits?

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -331,33 +352,112 @@ public final class KafkaConsumer: Sendable, Service {
...
             }
         }
     }
 
+    /// Read `maxMessages` consumer messages from Kafka.
+    ///
+    /// - Parameters:
+    ///     - client: Client used for handling the connection to the Kafka cluster.
+    ///     - maxMessages: Maximum amount of consumer messages to read in this invocation.
+    private func batchConsumerPoll(
```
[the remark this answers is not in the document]
FranzBusch: Let's track that in a follow up issue

Sources/Kafka/KafkaConsumer.swift
FranzBusch: Yeah we should probably set this to the difference of `high - low`.

Sources/Kafka/RDKafka/RDKafkaClient.swift
FranzBusch: I think there is going to be a clear split now between message related errors and run loop errors.

FranzBusch (approved):
Overall, this looks good to me now and I think we should go-ahead and merge it and then  see if we encounter any problems with it.

PR #155, opened 2023-12-04
Remove curl dependency and disable OAuth Bearer SASL mechanism

FranzBusch:
# Motivation
The curl dependency is just necessary for the OAuth Bearer SASL mechanism. Let's remove that configuration option for now.

# Modification
Removes the curl dependency.


Sources/Kafka/Configuration/KafkaConfiguration+Security.swift
```diff
@@ -309,7 +309,7 @@ extension KafkaConfiguration {
             }
         }
 
-        public struct OAuthBearerMethod: Sendable, Hashable {
+        struct OAuthBearerMethod: Sendable, Hashable {
```
blindspotbounty, line 312: Won't it be (formally) a breaking change if anyone uses that one for auth?

FranzBusch: Yes it would but we haven't released a major yet so removing APIs is still possible. I would love to get the package into a state where we don't require system dependencies at all.

blindspotbounty (approved):

mimischi:
I can see that there's lots of duplicated `#define` statements in `config.h`. I don't think that's expected, is it?

These

https://github.com/swift-server/swift-kafka-client/blob/ff579ee08a9140a8e1b796de04dcff7f78241370/Sources/Crdkafka/custom/config/config.h#L22-L31

are duplicated just a few lines below

https://github.com/swift-server/swift-kafka-client/blob/ff579ee08a9140a8e1b796de04dcff7f78241370/Sources/Crdkafka/custom/config/config.h#L34-L38

`ENABLE_LZ4_EXT` is even defined three times in the above two segments. There's some more duplication further down in the file.

mimischi:
This was confusing as heck, but I've got things to build locally again. Turns out `config.h` defines both `ENABLE_CURL` and `WITH_CURL`. The former is only referenced in [`configure.libcurl`](https://github.com/confluentinc/librdkafka/blob/267367c9475c2154e72eafe6ff1957518cb2ed1a/mklove/modules/configure.libcurl), while `WITH_CURL` is all over the place and controls the inclusion of `rdkafka_sasl_oauthbearer.c`—and that leads to build errors without `curl`.

mimischi:
Oh fun, the build is still broken on Linux

Package.swift
```diff
@@ -72,7 +78,6 @@ let package = Package(
                 .define("_GNU_SOURCE", to: "1"),  // Fix build error for Swift 5.9 onwards
             ],
             linkerSettings: [
-                .linkedLibrary("curl"),
                 .linkedLibrary("sasl2"),
```
FranzBusch, line 81: Should we remove this here as well then?

FranzBusch: Okay we should look into building libsassl with SwiftPM then to avoid the system dependency.

PR #157, opened 2023-12-06
Defer source.finish()

blindspotbounty:
Flush may throw exception thus source will not be finished.
Exception can be e.g. timeout (due to network outage): 
```
Caught exception KafkaError.rdKafkaError: Local: Timed out Kafka/RDKafkaClient.swift...
```


mr-swifter (approved):

mr-swifter (approved):

mr-swifter (changes requested):
Test

mr-swifter (approved):

PR #160, opened 2024-01-03
Commit messages by topic, partition, offset

omarkj:
Motivation:

In some cases, for instances when messages are handed out to other processes, it is not reasonable to hold on to the whole Kafka message to be able to commit it at a later time. It consumes much less memory to hold only onto the details needed to commit the message.

Modifications:

Add new public methods that allow users to commit messages if they have a reference to the topic, partition, and offset instead of only the whole message.

Result:

Developers no longer need to keep the whole message around to commit it, only the topic, partition, and offset.


Sources/Kafka/KafkaConsumer.swift
```diff
@@ -492,6 +492,27 @@ public final class KafkaConsumer: Sendable, Service {
...
+    ///
+    /// - Warning: This method fails if the ``KafkaConsumerConfiguration/isAutoCommitEnabled`` configuration property is set to `true` (default).
+    ///
+    /// - Parameters:
+    ///     - topic: Topic where the message that should be marked as read resides.
+    ///     - partition: Partition where the message that should be marked as read resides.
+    ///     - offset: Offset of the message that shall be marked as read.
+    /// - Throws: A ``KafkaError`` if committing failed.
+    public func scheduleCommit(topic: String,
+                               partition: KafkaPartition,
```
FranzBusch, line 514: Wondering if we have to expose both a topic and a partition argument here. Topic we should be able to get from our config and the consumer should always stick to the same topic. Partition is a bit more interesting due to rebalancing events.

NIT: Could you also fix the line breaks in this whole commit. We normally line break after the `(` and before `)`. This is matching what Xcode does.

omarkj: A client can be configured to consume from multiple topics and partitions (see `assign/1`) so I do not think there is a way around passing the topic in explicitly.

FranzBusch: Right now a client can only be configured to consume from a single topic since we are not publicly exposing `assign`. Following this reasoning we shouldn't expose `topic` here either but take it from the config,

FranzBusch (changes requested):

blindspotbounty:
@omarkj thank you for your PR! There are some notes.

Agree with @FranzBusch that indeed currently assign (`.partition`) does not allow to accept several topics. 
Though it doesn't lack sense to have topic as argument as there is other consumption strategy: `.group` that allows to subscribe to multiple topics by either explicitly specifying them, either subscribing to wildcard `.group(id: "<group id>", topics: ["topic1", "topic2", "^some_prefixed_topic_wildcard.*"])`. 

However, I would probably organise it as some dedicated struct, i.e. `TopicPartitionOffset`. That have couple of advantages now and for future use:
1. That would allow to copy this info in one short from consumer message
2. Maybe then migrate `.partition` consumption strategy to the same structure in future
3. The same structure can be used for rebalance implementation in future

For example: 
```swift
struct TopicPartitionOffset: Hashable, Equitable? {
    public let topic: String
    public let partition: KafkaPartition
    public let offset: KafkaOffset
    /// no public constructor (at least so far)
}

struct KafkaConsumerMessage {
     var topicPartitionOffset: TopicPartitionOffset { .init(topic: self.topic, partition: self.partition, offset: self.offset) }
}

final class KafkaConsumer {
...
    func commit(_ topicPartitionOffset) throws
}
```

The usage would be the following:
```
for try await msg in consumer.messages {
    consumer.commit(msg.topicPartitionOffset) // here or after some time consuming processing now only need to save topicPartitionOffset
}
```

PR #169, opened 2024-08-05
adopt swift6

blindspotbounty:


FranzBusch:
We should disable warnings as errors on 5.10. We won't be able to make that work probably.

PR #177, opened 2024-11-13
Enable Swift 6 support

mimischi:
Enable Swift 6 support. `String.init(validatingUTF8:)` is deprecated, and we use `String.init(validatingCString:)` instead.

Closes https://github.com/swift-server/swift-kafka-client/pull/169


Lukasa (changes requested):
We don’t intend to drop support for 5.9 or 5.10 at this time.

FranzBusch (approved):

Lukasa (approved):
Nice! :sparkles:

PR #178, opened 2024-11-14
Poll for messages using `TaskExecutor`

mimischi:
We currently sleep for `pollInterval` when no new messages have been polled from the cluster. This leads to unnecessary slowness of the client. Instead of doing that, we now break up the polling of messages into two distinct approaches:

1. Attempt to poll synchronously: if there a message is polled, we return it. If there is no message, we immediately go to step 2.
2. We create a `DispatchQueue` and run the `consumerPoll` on it using `withTaskExecutorPreference`. We make the `consumerPoll` call wait for up to `pollInterval` before bailing.

This prevents us from sleeping on the running thread, and frees up cycles to do other work if required.

Resolves https://github.com/swift-server/swift-kafka-client/issues/165

### DIFF
--- a/Sources/Kafka/KafkaConsumer.swift
+++ b/Sources/Kafka/KafkaConsumer.swift
@@ -12,6 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Dispatch
 import Logging
 import NIOConcurrencyHelpers
 import NIOCore
@@ -72,6 +73,7 @@ public struct KafkaConsumerMessages: Sendable, AsyncSequence {
     public struct AsyncIterator: AsyncIteratorProtocol {
         private let stateMachineHolder: MachineHolder
         let pollInterval: Duration
+        let queue: NaiveQueueExecutor
 
         private final class MachineHolder: Sendable {  // only for deinit
             let stateMachine: LockedMachine
@@ -88,21 +90,30 @@ public struct KafkaConsumerMessages: Sendable, AsyncSequence {
         init(stateMachine: LockedMachine, pollInterval: Duration) {
             self.stateMachineHolder = .init(stateMachine: stateMachine)
             self.pollInterval = pollInterval
+            self.queue = NaiveQueueExecutor(DispatchQueue(label: "com.swift-server.swift-kafka.message-consumer"))
         }
 
         public func next() async throws -> Element? {
-            // swift-kafka-client issue: https://github.com/swift-server/swift-kafka-client/issues/165
-            // Currently use Task.sleep() if no new messages, should use task executor preference when implemented:
-            // https://github.com/apple/swift-evolution/blob/main/proposals/0417-task-executor-preference.md
             while !Task.isCancelled {
                 let action = self.stateMachineHolder.stateMachine.withLockedValue { $0.nextConsumerPollLoopAction() }
 
                 switch action {
                 case .poll(let client):
-                    if let message = try client.consumerPoll() {  // non-blocking call
+                    // Attempt to fetch a message synchronously. Bail
+                    // immediately if no message is waiting for us.
+                    if let message = try client.consumerPoll() {
                         return message
                     }
-                    try await Task.sleep(for: self.pollInterval)
+
+                    #if swift(>=6.0)
+                    // Wait on a separate thread for the next message.
+                    return try await withTaskExecutorPreference(queue) {
+                        try client.consumerPoll(for: Int32(self.pollInterval.inMilliseconds))
+                    }
+                    #else
+                    // No messages. Sleep a little.
+                    return try await Task.sleep(for: self.pollInterval)
+                    #endif
                 case .suspendPollLoop:
                     try await Task.sleep(for: self.pollInterval)  // not started yet
                 case .terminatePollLoop:
--- a/Sources/Kafka/KafkaConsumer.swift
+++ b/Sources/Kafka/KafkaConsumer.swift
@@ -73,7 +73,9 @@ public struct KafkaConsumerMessages: Sendable, AsyncSequence {
     public struct AsyncIterator: AsyncIteratorProtocol {
         private let stateMachineHolder: MachineHolder
         let pollInterval: Duration
+        #if swift(>=6.0)
         let queue: NaiveQueueExecutor
+        #endif
 
         private final class MachineHolder: Sendable {  // only for deinit
             let stateMachine: LockedMachine
@@ -90,7 +92,9 @@ public struct KafkaConsumerMessages: Sendable, AsyncSequence {
         init(stateMachine: LockedMachine, pollInterval: Duration) {
             self.stateMachineHolder = .init(stateMachine: stateMachine)
             self.pollInterval = pollInterval
+            #if swift(>=6.0)
             self.queue = NaiveQueueExecutor(DispatchQueue(label: "com.swift-server.swift-kafka.message-consumer"))
+            #endif
         }
 
         public func next() async throws -> Element? {
--- a/Sources/Kafka/KafkaConsumer.swift
+++ b/Sources/Kafka/KafkaConsumer.swift
@@ -12,7 +12,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-import Dispatch
 import Logging
 import NIOConcurrencyHelpers
 import NIOCore
@@ -74,7 +73,7 @@ public struct KafkaConsumerMessages: Sendable, AsyncSequence {
         private let stateMachineHolder: MachineHolder
         let pollInterval: Duration
         #if swift(>=6.0)
-        let queue: NaiveQueueExecutor
+        private let queue: DispatchQueueTaskExecutor
         #endif
 
         private final class MachineHolder: Sendable {  // only for deinit
@@ -93,7 +92,9 @@ public struct KafkaConsumerMessages: Sendable, AsyncSequence {
             self.stateMachineHolder = .init(stateMachine: stateMachine)
             self.pollInterval = pollInterval
             #if swift(>=6.0)
-            self.queue = NaiveQueueExecutor(DispatchQueue(label: "com.swift-server.swift-kafka.message-consumer"))
+            self.queue = DispatchQueueTaskExecutor(
+                DispatchQueue(label: "com.swift-server.swift-kafka.message-consumer")
+            )
             #endif
         }
 
@@ -111,6 +112,7 @@ public struct KafkaConsumerMessages: Sendable, AsyncSequence {
 
                     #if swift(>=6.0)
                     // Wait on a separate thread for the next message.
+                    // The call below will block for `pollInterval`.
                     return try await withTaskExecutorPreference(queue) {
                         try client.consumerPoll(for: Int32(self.pollInterval.inMilliseconds))
                     }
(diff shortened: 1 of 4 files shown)


mimischi:
Haven't done any benchmarks just yet. Would like to do some, before we go about merging this.

mimischi:
The benchmark looks better, but the benchmark suite says all differences are negative—and it's setting a positive difference as the threshold, so while this PR is better, the benchmark ends up failing?

```

=====================================================================================================
Threshold deviations for SwiftKafkaConsumerBenchmarks:SwiftKafkaConsumer_basic_consumer_messages_1000
=====================================================================================================
╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Time (wall clock) (ms, %)                │            main │              PR │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p90                                      │             642 │             149 │             -76 │              35 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Time (total CPU) (ms, %)                 │            main │              PR │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p90                                      │              30 │              17 │             -42 │              35 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Throughput (# / s) (#, %)                │            main │              PR │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p90                                      │               2 │               7 │            -250 │              35 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Context switches (#, %)                  │            main │              PR │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p90                                      │            1230 │             392 │             -68 │              35 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ (Alloc + Retain) - Release Δ (#, %)      │            main │              PR │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p90                                      │            2004 │            1223 │             -39 │              20 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Releases (K, %)                          │            main │              PR │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p90                                      │              14 │              11 │             -22 │              20 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

=========================================================================================================
Threshold deviations for SwiftKafkaConsumerBenchmarks:SwiftKafkaConsumer_with_offset_commit_messages_1000
=========================================================================================================
╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Time (wall clock) (ms, %)                │            main │              PR │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p90                                      │             635 │             119 │             -81 │              35 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Time (total CPU) (ms, %)                 │            main │              PR │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p90                                      │              38 │              15 │             -62 │              35 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Throughput (# / s) (#, %)                │            main │              PR │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p90                                      │               2 │               8 │            -300 │              35 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ Context switches (#, %)                  │            main │              PR │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p90                                      │            1432 │             507 │             -64 │              35 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

╒══════════════════════════════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╕
│ (Alloc + Retain) - Release Δ (#, %)      │            main │              PR │    Difference % │     Threshold % │
╞══════════════════════════════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╡
│ p90                                      │            1994 │            1186 │             -40 │              20 │
╘══════════════════════════════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╛

New baseline 'PR' is BETTER than the 'main' baseline thresholds.

error: benchmarkThresholdImprovement
Retcode is 1
Benchmark failed
```

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -12,6 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Dispatch
```
FranzBusch, line 15: Don't think we need this import here

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -88,21 +92,32 @@ public struct KafkaConsumerMessages: Sendable, AsyncSequence {
...
                 case .poll(let client):
-                    if let message = try client.consumerPoll() {  // non-blocking call
+                    // Attempt to fetch a message synchronously. Bail
+                    // immediately if no message is waiting for us.
+                    if let message = try client.consumerPoll() {
                         return message
                     }
+
+                    #if swift(>=6.0)
+                    // Wait on a separate thread for the next message.
```
FranzBusch, line 113: Can we add here that the call below is blocking here.

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -88,21 +92,32 @@ public struct KafkaConsumerMessages: Sendable, AsyncSequence {
...
+                    return try await withTaskExecutorPreference(queue) {
+                        try client.consumerPoll(for: Int32(self.pollInterval.inMilliseconds))
```
FranzBusch, line 115: What happens after the time out?

Sources/Kafka/Utilities/NaiveQueueExecutor.swift
```diff
@@ -0,0 +1,39 @@
...
+// See CONTRIBUTORS.txt for the list of swift-kafka-client project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=6.0)
+import Dispatch
+
+final class NaiveQueueExecutor: TaskExecutor {
```
FranzBusch, line 18:
```suggestion
final class DispatchQueueTaskExecutor: TaskExecutor {
```

Sources/Kafka/Utilities/NaiveQueueExecutor.swift
```diff
@@ -0,0 +1,39 @@
...
+    let queue: DispatchQueue
```
FranzBusch, line 19:
```suggestion
   private let queue: DispatchQueue
```

Sources/Kafka/Utilities/NaiveQueueExecutor.swift
```diff
@@ -0,0 +1,39 @@
...
+
+    init(_ queue: DispatchQueue) {
+        self.queue = queue
+    }
+
+    public func enqueue(_ _job: consuming ExecutorJob) {
+        let job = UnownedJob(_job)
+        queue.async {
+            job.runSynchronously(
```
FranzBusch, line 28: @ktoso Should we call this `runSynchronously` or the one that also takes `isolatedTo`?

FranzBusch:
> The benchmark looks better, but the benchmark suite says all differences are negative—and it's setting a positive difference as the threshold, so while this PR is better, the benchmark ends up failing?

Yes the benchmarks improved significantly! Negative numbers means fewer allocations.

mimischi:
> Negative numbers means fewer allocations.

It's just funny that the benchmarking suite thinks the benchmark failed, because we are below the positive threshold with our negative numbers :)

Sources/Kafka/KafkaConsumer.swift
mimischi: Whoops. Left over from a refactor. Is there any linter that can help point out unused imports?

mimischi:
Also, should this PR—once approved—update the benchmark baseline as well?

hassila:
> > Negative numbers means fewer allocations.
> 
> 
> 
> It's just funny that the benchmarking suite thinks the benchmark failed, because we are below the positive threshold with our negative numbers :)

There are different return codes so CI can choose how to handle that - for an expected improvement one would check in new baselines manually anyway - for an unexpected  improvement something may be wrong with eg. The benchmark setup. 

But super nice improvements here 👍🏻

mimischi:
@hassila Oh, fair enough. I've not thought about the unexpected improvement situation. Thanks for bringing that up!

Sources/Kafka/Utilities/DispatchQueueTaskExecutor.swift
```diff
@@ -0,0 +1,39 @@
...
+
+    init(_ queue: DispatchQueue) {
+        self.queue = queue
+    }
+
+    public func enqueue(_ _job: consuming ExecutorJob) {
+        let job = UnownedJob(_job)
+        queue.async {
+            job.runSynchronously(
+                on: self.asUnownedTaskExecutor()
```
ktoso, line 29: if you're able to use the 

```
public func runSynchronously(isolatedTo serialExecutor: UnownedSerialExecutor,
                             taskExecutor: UnownedTaskExecutor) {
```


here and pass the queue's `UnownedSerialExecutor` that would be preferable as it would AFAIR be more correct in tracking where this is isolated to.

If this is a pain because the queue does not conform to SerialExecutor on some platforms still... then perhaps conditionalize it to platforms where it is? Or leave as is and let me know and we'll chase fixing the conformance.

ktoso (approved):
This looks good to me, check the run func but either way I think this is looking good

Sources/Kafka/Utilities/DispatchQueueTaskExecutor.swift
FranzBusch: Got you. Since we own the underlying queue here is owned by us nothing should be isolated to it.

PR #179, opened 2024-11-15
Add headers to `KafkaAcknowledgedMessage`

malletgu:
When a message is being sent to Kafka the acknowledgement should also contains the headers.

### Motivation:

Resolves https://github.com/swift-server/swift-kafka-client/issues/144

### Modifications:

Reused the parsing function from KafkaConsumer in the producer path to return headers in the ack.

### Result:

Headers available in the acknowledgement for producers and still available for consumers as well.


FranzBusch (approved):
Nice! Thanks for working on this

PR #197, opened 2025-06-29
Add support for `fetch.wait.max.ms` Kafka consumer config

supersonicbyte:
Motivation:

Currently the API doesn't provide a way to set the `fetch.wait.max.ms` Kafka consumer config. This PR enables the user to set it.

Modifications:

Added support for `fetch.wait.max.ms` inside of `KafkaConsumerConfiguration`.

Result:

We can set `fetch.wait.max.msa`.


supersonicbyte:
@FranzBusch can we merge this please?

PR #221, opened 2026-04-06
Add callback-based rebalance handling and offload blocking C calls

ssikka100:
## Motivation

1. **No rebalance event delivery**: `rd_kafka_conf_set_rebalance_cb` was never called, so consumers received no notification of partition assignment or revocation. Without this, applications cannot implement commit-on-revoke, flush in-memory state on partition loss, or clean up per-partition resources during consumer group rebalancing. All major Kafka clients (Java, Rust, Python, Go) expose rebalance callbacks.

2. **Blocking C calls on cooperative thread pool**: `committed()` and `seekPartitions()` called blocking C functions (`rd_kafka_committed`, `rd_kafka_seek_partitions`) directly on Swift Concurrency's cooperative thread pool. If the broker is slow or unreachable, these calls block for the full timeout (default 5 seconds), preventing `Task.isCancelled` checks, starving other async tasks, and causing `.timeLimit` to fail to enforce timeouts.

## Modifications

- Add `RebalanceContext` class with C rebalance callback that handles assign/unassign synchronously inside `rd_kafka_consumer_poll()`, supporting both **eager** (`rd_kafka_assign`) and **cooperative** (`rd_kafka_incremental_assign`/`rd_kafka_incremental_unassign`) protocols. Rebalance events are buffered in a thread-safe `NIOLockedValueBox` and drained by the Swift event loop.
- Add `.rebalance` case in `consumerEventPoll` to handle shutdown revoke events that arrive on the main queue (after `rd_kafka_consumer_close_queue` forwards the consumer queue). Both dispatch paths (C callback during normal operation, event poll during shutdown) route through the same `RebalanceContext.handleRebalance` method.
- Add `KafkaConsumerRebalance` public struct with `Kind` enum (`.assign`, `.revoke`, `.error`) and `KafkaConsumerEvent.rebalance` case for user-facing event delivery through `KafkaConsumerEvents` async sequence.
- Store `RebalanceContext` in state machine alongside `RDKafkaClient` in `.initializing`/`.running`/`.finishing` states to prevent use-after-free if the message iterator outlives `KafkaConsumer`.
- Add `defer { eventsSource?.finish() }` in `eventRunLoop` to guarantee the events stream ends on all exit paths (normal, cancellation, error).
- Add `finalDrain()` method called on `.terminatePollLoop` to poll+drain one last time after `isConsumerClosed` returns true, closing the race window where the close process completes before the event loop polls the shutdown revoke.
- Offload `committed()` and `seekPartitions()` to a shared concurrent `DispatchQueue` via `withCheckedThrowingContinuation`, matching the existing `flush()` pattern.
- Move `flush()` to the same shared concurrent `DispatchQueue`, eliminating per-call queue allocation.
- Add `RDKafkaTopicPartitionList` `@unchecked Sendable` conformance with safety documentation for capture in `@Sendable` closures.
- Remove dead `KafkaConsumerEvent.init(_ event:)` that always `fatalError`'d.

## API Changes

Two public method signatures changed from `throws` to `async throws`:

| Method | Before | After |
|--------|--------|-------|
| `KafkaConsumer.committed(topicPartitions:timeout:)` | `throws -> [KafkaTopicPartitionOffset]` | `async throws -> [KafkaTopicPartitionOffset]` |
| `KafkaConsumer.seek(topicPartitionOffsets:timeout:)` | `throws` | `async throws` |

**Note**: Both methods were added in PR #217 (merged to `main` but not yet included in any tagged release — the latest release is `1.0.0-alpha.9`). This signature change has zero migration impact for downstream users on a released version.

**Why `async throws`**: Both methods call blocking C functions that can block for up to `timeout` (default 5 seconds) waiting for a broker response. The `async throws` signature ensures the Swift task suspends at `await` while the blocking call runs on a separate GCD thread, freeing the cooperative thread pool. This matches the pattern already used by `flush()`.

**Migration** (for code on `main`): Add `await` at call sites:
```swift
// Before:
let offsets = try consumer.committed(topicPartitions: partitions)
try consumer.seek(topicPartitionOffsets: targets)

// After:
let offsets = try await consumer.committed(topicPartitions: partitions)
try await consumer.seek(topicPartitionOffsets: targets)
```

## Result

Consumers receive rebalance events (`.assign`, `.revoke`, `.error`) through the `KafkaConsumerEvents` async sequence. Both eager and cooperative rebalance protocols are supported. The library handles assign/unassign automatically — events are informational for application-level bookkeeping. Blocking C calls no longer starve the cooperative thread pool. Graceful shutdown completes cleanly with all events drained. At-least-once delivery semantics preserved across rebalances.

## Tests & Verification

- 41 unit tests (RebalanceContext buffer mechanics with concurrent inject+drain, KafkaConsumerRebalance equality/hashable/error, closed-consumer guards for async committed/seek)
- 19 integration tests including 5 new: rebalance event delivery, concurrent message+rebalance flow, cooperative protocol, shutdown revoke delivery, at-least-once with no message loss across rebalance
- All existing tests updated for async `committed()`/`seek()` signatures
- Successful completion of new/existing tests via `make docker-test`

### DIFF
--- a/Sources/Kafka/RDKafka/RebalanceContext.swift
+++ b/Sources/Kafka/RDKafka/RebalanceContext.swift
@@ -1,0 +1,268 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the swift-kafka-client open source project
+//
+// Copyright (c) 2026 Apple Inc. and the swift-kafka-client project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of swift-kafka-client project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crdkafka
+import Logging
+import NIOConcurrencyHelpers
+
+/// C-compatible rebalance callback passed to `rd_kafka_conf_set_rebalance_cb`.
+///
+/// Invoked **synchronously** inside `rd_kafka_consumer_poll()` when a consumer
+/// group rebalance occurs. The callback must call `rd_kafka_assign` /
+/// `rd_kafka_incremental_assign` / `rd_kafka_incremental_unassign` before
+/// returning — librdkafka blocks the consumer group protocol until it does.
+///
+/// The `opaque` parameter is the value set via `rd_kafka_conf_set_opaque`,
+/// which we set to an `Unmanaged<RebalanceContext>`.
+///
+/// - Important: The parameter types must exactly match the C import of
+///   `rd_kafka_conf_set_rebalance_cb`'s function pointer typedef.
+private func rebalanceCallback(
+    _ rk: OpaquePointer?,
+    _ err: rd_kafka_resp_err_t,
+    _ partitions: UnsafeMutablePointer<rd_kafka_topic_partition_list_t>?,
+    _ opaque: UnsafeMutableRawPointer?
+) {
+    guard let opaque else { return }
+    let context = Unmanaged<RebalanceContext>.fromOpaque(opaque).takeUnretainedValue()
+    context.handleRebalance(kafkaHandle: rk, error: err, partitions: partitions)
+}
+
+/// Holds the state needed by the C rebalance callback.
+///
+/// This object is set as the config opaque pointer via `rd_kafka_conf_set_opaque`
+/// and lives for the lifetime of the `KafkaConsumer`. The C callback receives it
+/// as the `void* opaque` parameter and calls ``handleRebalance(kafkaHandle:error:partitions:)``.
+///
+/// Marked `@unchecked Sendable` because all mutable state (`pendingEvents`) is
+/// protected by `NIOLockedValueBox`. The class must be a reference type because
+/// it is passed to C via `Unmanaged` as the config opaque pointer.
+///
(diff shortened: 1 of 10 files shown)


Sources/Kafka/RDKafka/RebalanceContext.swift
```diff
@@ -0,0 +1,268 @@
...
+///
+/// Marked `@unchecked Sendable` because all mutable state (`pendingEvents`) is
+/// protected by `NIOLockedValueBox`. The class must be a reference type because
+/// it is passed to C via `Unmanaged` as the config opaque pointer.
+///
+/// Thread safety: The rebalance callback fires synchronously inside
+/// `rd_kafka_consumer_poll()`, which in our architecture runs on either the
+/// cooperative thread pool or a GCD thread. The pending events buffer is
+/// protected by `NIOLockedValueBox`. The `logger` is `Sendable`.
+final class RebalanceContext: @unchecked Sendable {
```
FranzBusch, line 56: This should not require an `@unchecked Sendable` since `NIOLockedValueBox` is already Sendable. Instead of using `NIOLockedValueBox` we should start using `Mutex` instead here. But it is fine to do that migration in a follow up PR.

FranzBusch (approved):

PR #222, opened 2026-04-07
Add awaitable producer send (sendAndAwait)

ssikka100:
## Motivation

The current `send()` is fire-and-forget — it returns a `KafkaProducerMessageID` immediately without waiting for broker acknowledgment. To confirm delivery, callers must create a producer with `makeProducerWithEvents`, run a separate task iterating the `KafkaProducerEvents` async sequence, and match each `KafkaDeliveryReport.id` back to the original message ID.

This is significantly more complex than equivalent APIs in other Kafka clients:
- **Java**: `producer.send(record).get()` — single blocking call
- **Rust**: `producer.send(record, timeout).await` — single awaitable call

## Modifications

- Add `sendAndAwait(_:)` public method on `KafkaProducer` that suspends until the broker acknowledges or rejects the message, returning a `KafkaDeliveryReport`
- Model pending continuations as a tri-state (`ContinuationState`) inside the lock-protected state machine to handle the `withTaskCancellationHandler` race where `onCancel` can fire before the operation body:
  - `.initialized` — slot allocated, continuation not yet registered
  - `.pending(continuation)` — continuation registered, waiting for delivery report
  - `.cancelled` — `onCancel` fired before registration, `registerContinuation` rejects immediately
- Add `failPendingContinuations()` called on both shutdown paths (`flushFinishSourceAndTerminatePollLoop` and `terminatePollLoop`) to resume pending continuations with `connectionClosed` error, preventing `sendAndAwait` from hanging on producer shutdown
- Add `.deliveryReport` to `subscribedEvents` in the no-events producer `init(config:logger:)` so delivery reports appear on the event queue — required for `sendAndAwait` to work without `makeProducerWithEvents`
- Modify `_run()` event loop: `.pollWithoutYield` now processes delivery reports for continuations instead of discarding them. `.pollAndYield` resumes continuations first, then yields all reports to the events sequence (complete log, no filtering)
- Add explicit `init(logger:)` to `StateMachine` (private `pendingContinuations` suppressed the memberwise init)

## API

**New method** on `KafkaProducer`:

```swift
public func sendAndAwait<Key, Value>(
    _ message: KafkaProducerMessage<Key, Value>
) async throws -> KafkaDeliveryReport
```

**Backward compatible**: `send()` unchanged. Both can be used on the same producer — `sendAndAwait` messages get individual continuations, `send` messages get batched delivery reports via the events sequence. All delivery reports flow to the events sequence regardless of how the message was sent.

**Cancellation**: If the caller's task is cancelled, `sendAndAwait` resumes with `CancellationError`. The message may still be delivered by librdkafka, but the caller exits immediately. The delivery report still appears in the events sequence.

**Shutdown**: If the producer shuts down before a delivery report arrives, `sendAndAwait` resumes with `KafkaError.connectionClosed`.

## Result

Producers support both fire-and-forget (`send`) and awaitable (`sendAndAwait`) patterns. Task cancellation and producer shutdown are handled without leaking continuations. Existing behavior is unchanged.

### Unit Tests (`Tests/KafkaTests/KafkaProducerTests.swift`)

| Test | What It Verifies |
|------|-----------------|
| `sendAndAwaitFailsOnClosedProducer` | Throws `KafkaError` after graceful shutdown |
| `sendAndAwaitCancellation` | Task cancellation resumes with `CancellationError` |
| `sendAndAwaitFailsOnShutdown` | Shutdown before delivery resumes with `connectionClosed` |

### Integration Tests (`Tests/IntegrationTests/KafkaTests.swift`)

| Test | What It Verifies |
|------|-----------------|
| `sendAndAwaitReturnsDeliveryReport` | Single message — acknowledged with topic/partition/offset |
| `sendAndAwaitMultipleMessages` | 5 sequential messages — all acknowledged |
| `sendAndAwaitWorksWithoutEventsSequence` | Works on `KafkaProducer(config:logger:)` without events |
| `sendAndAwaitConcurrentSends` | 10 concurrent sends via TaskGroup — all acknowledged |
| `sendAndSendAndAwaitMixedOnSameProducer` | Both `send()` and `sendAndAwait()` reports appear in events sequence |


Sources/Kafka/KafkaProducer.swift
```diff
@@ -235,16 +235,27 @@ public final class KafkaProducer: Service, Sendable {
...
                 for event in events {
                     switch event {
                     case .statistics(let statistics):
                         self.config.metrics.update(with: statistics)
                     case .deliveryReport(let reports):
-                        _ = source?.yield(.deliveryReports(reports))
+                        let unclaimed = self.resumeContinuations(for: reports)
+                        if !unclaimed.isEmpty {
+                            _ = source?.yield(.deliveryReports(unclaimed))
+                        }
```
FranzBusch: Should we really not send the delivery reports for the messages that are awaited to the events source?

Sources/Kafka/KafkaProducer.swift
```diff
@@ -295,6 +332,52 @@ public final class KafkaProducer: Service, Sendable {
...
+    /// This provides a Java-like `producer.send(record).get()` or Rust-like
+    /// `producer.send(record).await` experience — a single awaitable call that confirms delivery.
+    ///
+    /// - Parameter message: The ``KafkaProducerMessage`` to send.
+    /// - Returns: A ``KafkaDeliveryReport`` with the delivery status.
+    /// - Throws: A ``KafkaError`` if the message could not be enqueued or was rejected by the broker.
+    public func sendAndAwait<Key, Value>(
+        _ message: KafkaProducerMessage<Key, Value>
+    ) async throws -> KafkaDeliveryReport {
+        try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<KafkaDeliveryReport, Error>) in
```
FranzBusch: We need to handle cancellation here. What if the user's task is cancelled while we are awaiting? We should resume the continuation with a `CancellationError` most likely.

Sources/Kafka/KafkaProducer.swift
ssikka100: Good catch, fixed. All delivery reports now go to the events source regardless of whether a `sendAndAwait` continuation is also resumed. The events sequence is now a complete log of all delivery reports. `resumeContinuations` no longer filters reports,  it resumes matching continuations and reports are yielded to the source.

Sources/Kafka/KafkaProducer.swift
ssikka100: Agreed,  the original implementation was missing cancellation handling entirely. Updated to use `withTaskCancellationHandler`: when the user's task is cancelled, the `onCancel` closure removes the pending continuation from the dictionary and resumes it with `CancellationError()`. The message itself is already enqueued in librdkafka at that point and will still be delivered to the broker, but the caller's task exits immediately. The delivery report will still appear in the events sequence for any monitoring/auditing code that consumes it.

Sources/Kafka/KafkaProducer.swift
```diff
@@ -295,6 +327,68 @@ public final class KafkaProducer: Service, Sendable {
...
+                }
+            }
+        } onCancel: {
+            // Task was cancelled — remove the pending continuation and resume with CancellationError.
+            // The message is already enqueued in librdkafka and will be delivered regardless,
+            // but the caller no longer cares about the result.
+            let cancelled = self.stateMachine.withLockedValue {
+                $0.removeContinuation(for: newMessageID)
+            }
+            cancelled?.resume(throwing: CancellationError())
```
FranzBusch, line 420: There is a race condition here. The `onCancel` can be called before the operation body above if the task is already cancelled when calling `withTaskCancellationHandler`. You need to handle in your state machine the case where that happens

ssikka100: added a Task.isCancelled check right after registerContinuation to handle this. If onCancel fired first, the flag is already true so we clean up immediately. removeContinuation returning nil prevents any double-resume.

FranzBusch: The check at the top is not enough. We still have a race where after the check the task can get canceled. The other checks are similarly not enough. You need to handle the reality that when you try to store the continuation into the mutex that the task is cancelled. We need to model the tri-state inside the lock protected state.:
- No continuation
- Continuation
- Cancelled

ssikka100: - Refactored the continuation as a tri-state (initialized / pending / cancelled) inside the lock-protected state machine. 
- initializeContinuation allocates the slot before entering withTaskCancellationHandler. If onCancel fires first, it sees .initialized and leaves a .cancelled tombstone. registerContinuation checks the slot, if .cancelled, returns .alreadyCancelled and the continuation is resumed with CancellationError immediately. 
- Also added failPendingContinuations on both shutdown paths to prevent leaked continuations on producer shutdown.

Sources/Kafka/KafkaProducer.swift
```diff
@@ -321,6 +347,82 @@ public final class KafkaProducer: Service, Sendable {
...
+    ///
+    /// - Parameter message: The ``KafkaProducerMessage`` to send.
+    /// - Returns: A ``KafkaDeliveryReport`` with the delivery status.
+    /// - Throws: A ``KafkaError`` if the message could not be enqueued or was rejected by the broker.
+    public func sendAndAwait<Key, Value>(
+        _ message: KafkaProducerMessage<Key, Value>
+    ) async throws -> KafkaDeliveryReport {
+        // if the task is already cancelled, bail out immediately
+        // without enqueuing the message in librdkafka.
+        try Task.checkCancellation()
```
FranzBusch: We don't need this check here anymore. Since we are handling the cancellation in the continuation anyways.

Sources/Kafka/KafkaProducer.swift
```diff
@@ -321,6 +347,82 @@ public final class KafkaProducer: Service, Sendable {
...
+                (continuation: CheckedContinuation<KafkaDeliveryReport, Error>) in
+                // Transition from .initialized to .pending BEFORE producing.
+                // The delivery report could arrive before produce() returns.
+                let result = self.stateMachine.withLockedValue {
+                    $0.registerContinuation(continuation, for: newMessageID)
+                }
+
+                // If the task was cancelled before we registered, the state machine
+                // already transitioned to .cancelled. Resume immediately.
+                if case .alreadyCancelled = result {
```
FranzBusch: Can we use a switch here so we are not missing any new results that we are potentially adding in the future.

Sources/Kafka/KafkaProducer.swift
```diff
@@ -515,5 +640,111 @@ extension KafkaProducer {
...
+        ///
+        /// Transitions the slot from `.initialized` to `.pending(continuation)`.
+        /// Returns `.alreadyCancelled` if `onCancel` already set the slot to `.cancelled`.
+        mutating func registerContinuation(
+            _ continuation: CheckedContinuation<KafkaDeliveryReport, Error>,
+            for messageID: UInt
+        ) -> RegisterContinuationResult {
+            guard let existing = self.pendingContinuations[messageID] else {
+                self.pendingContinuations[messageID] = .pending(continuation)
+                return .registered
```
FranzBusch: This can't happen right? We are always calling `initializeContinuation` before.

Sources/Kafka/KafkaProducer.swift
ssikka100: Done

Sources/Kafka/KafkaProducer.swift
ssikka100: Done

Sources/Kafka/KafkaProducer.swift
ssikka100: Changed to a fatalError to enforce the invariant.

FranzBusch (approved):

PR #224, opened 2026-04-14
Use in-process mock brokers for unit tests

ssikka100:
## Motivation

Unit tests create real librdkafka clients to test producer/consumer lifecycle, metrics, and state machine behavior. These tests relied on `localhost:9092` as the default bootstrap broker, causing librdkafka to attempt real TCP connections during test runs — an external dependency that unit tests should never have. Without a running broker, librdkafka generated connection errors that polluted test output.

librdkafka provides a `test.mock.num.brokers` configuration property that creates in-process mock brokers — supporting producers, consumers, consumer groups with offset commits, and topic auto-creation — without any external dependencies. This is the same approach used by Rust rdkafka for unit testing, where unit tests never set `bootstrap.servers`.

## Summary

- Add `useMockBroker(count:)` extensions on `KafkaProducerConfig` and `KafkaConsumerConfig` that set librdkafka's `test.mock.num.brokers` to create in-process mock brokers for unit testing
- Add `internal var additionalConfig: [String: String]` on `KafkaProducerConfig` and `KafkaConsumerConfig` for passing arbitrary librdkafka configuration properties
- Replace `localhost:9092` defaults in `KafkaProducerTests`, `KafkaConsumerTests`, and `KafkaMetricsTests` with `config.useMockBroker()`

## Test plan
- [x] `swift build --build-tests -Xswiftc -warnings-as-errors` passes
- [x] `swift test --skip IntegrationTests` passes — no broker connection attempts
- [x] `make docker-test` passes — integration tests unchanged

### DIFF
--- a/Tests/KafkaTests/MockCluster.swift
+++ b/Tests/KafkaTests/MockCluster.swift
@@ -14,19 +14,21 @@
 
 @testable import Kafka
 
-/// Helpers for configuring librdkafka's built-in mock cluster.
+/// Extensions for configuring librdkafka's built-in mock cluster.
 ///
 /// Setting `test.mock.num.brokers` causes librdkafka to create an in-process
 /// mock Kafka cluster, eliminating the need for a real broker in unit tests.
-enum MockBrokerConfig {
 
-    /// Apply mock broker configuration to a ``KafkaProducerConfig``.
-    static func apply(to config: inout KafkaProducerConfig, brokerCount: Int) {
-        config.additionalConfig["test.mock.num.brokers"] = "\(brokerCount)"
+extension KafkaProducerConfig {
+    /// Configure this producer to use librdkafka's in-process mock broker.
+    mutating func useMockBroker(count: Int = 1) {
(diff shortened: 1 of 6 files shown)


Tests/KafkaTests/MockCluster.swift
```diff
@@ -0,0 +1,32 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+@testable import Kafka
+
+/// Helpers for configuring librdkafka's built-in mock cluster.
+///
+/// Setting `test.mock.num.brokers` causes librdkafka to create an in-process
+/// mock Kafka cluster, eliminating the need for a real broker in unit tests.
+enum MockBrokerConfig {
```
FranzBusch: Instead of using an enum here can we instead write an extension to `KafkaProducerConfig` that allows us to set these values. That is the more Swifty way of doing this.

ssikka100: done

FranzBusch (approved):

PR #225, opened 2026-04-14
Register error events with typed error codes and surface through consumer/producer event sequences

ssikka100:
## Motivation

librdkafka reports errors (authentication failure, broker disconnection, SSL errors) through `RD_KAFKA_EVENT_ERROR`. The Swift client never subscribed to this event type — errors hit `default: break` and were silently dropped. A disconnected client gave no signal to the application.

Additionally, `KafkaError` discarded the original `rd_kafka_resp_err_t` code, making all librdkafka errors compare equal via `Equatable` and forcing applications to string-match error messages.

## Summary

- Subscribe to `RD_KAFKA_EVENT_ERROR` unconditionally in all init paths
- Add `.error(KafkaError)` to `KafkaConsumerEvent` and `KafkaProducerEvent`
- Add `handleErrorEvent()` on `RDKafkaClient` with NULL-safe reason extraction
- Handle errors in consumer `eventRunLoop()`, `finalDrain()`, and both producer poll branches — always log, yield to events source if available
- Add `logger` to `KafkaProducer` (matching consumer pattern)
- Add `KafkaError.RDKafkaCode` nested struct with typed constants (`allBrokersDown`, `authentication`, `transport`, `timedOut`, `ssl`, `fatal`, etc.)
- Add `rdKafkaCode` property on `KafkaError` — preserves the raw code (nil for non-librdkafka errors)
- Fix `Equatable`/`Hashable` to include `rdKafkaCode`

## Test plan
- [x] Build + unit tests pass

### DIFF
--- a/Sources/Kafka/KafkaConsumerEvent.swift
+++ b/Sources/Kafka/KafkaConsumerEvent.swift
@@ -22,6 +22,9 @@ public enum KafkaConsumerEvent: Sendable, Hashable {
     /// or initializing state on assign.
     case rebalance(KafkaConsumerRebalance)
 
+    /// An error reported by the Kafka client (e.g., broker disconnection, authentication failure).
+    case error(KafkaError)
+
     /// - Important: Always provide a `default` case when switching over this `enum`.
     case DO_NOT_SWITCH_OVER_THIS_EXHAUSITVELY
 }
--- a/Sources/Kafka/KafkaError.swift
+++ b/Sources/Kafka/KafkaError.swift
@@ -14,59 +14,61 @@
 
 import Crdkafka
 
-/// A type-safe wrapper around librdkafka's `rd_kafka_resp_err_t` error codes.
-///
-/// Use the static constants (e.g., `.allBrokersDown`, `.authentication`) to match
-/// against the ``KafkaError/rdKafkaErrorCode`` property without importing Crdkafka.
-public struct RDKafkaErrorCode: Hashable, Sendable, CustomStringConvertible {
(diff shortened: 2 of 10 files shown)


Sources/Kafka/KafkaError.swift
```diff
@@ -14,9 +14,57 @@
 
 import Crdkafka
 
+/// A type-safe wrapper around librdkafka's `rd_kafka_resp_err_t` error codes.
+///
+/// Use the static constants (e.g., `.allBrokersDown`, `.authentication`) to match
+/// against the ``KafkaError/rdKafkaErrorCode`` property without importing Crdkafka.
+public struct RDKafkaErrorCode: Hashable, Sendable, CustomStringConvertible {
```
FranzBusch: Can we nest this in the `KafkaError` type please.

Sources/Kafka/KafkaConsumerEvent.swift
FranzBusch, line 16: We should mark this enum as `@nonexhaustive` before we ship the real 1.0 but this requires 6.2 so let's track this in an issue and do it right before the release.

Sources/Kafka/KafkaError.swift
ssikka100: Done

Sources/Kafka/KafkaConsumerEvent.swift
ssikka100: Tracked here [#226 ]

FranzBusch (approved):

PR #227, opened 2026-04-20
Add dynamic subscribe, unsubscribe APIs to KafkaConsumer

ssikka100:
## Motivation

Topics are locked at initialization time via `consumptionStrategy` with no way to change subscriptions at runtime. The existing private `subscribe()` crashes on re-subscribe (`fatalError`), and `consumptionStrategy` is force-unwrapped in `_run()` which crashes if nil.

Applications that need dynamic topic routing, multi-tenant consumption, or config-driven subscriptions currently cannot use this client without recreating the consumer. This aligns the Swift client with the standard Kafka client contract where subscribe/unsubscribe are runtime operations, not initialization-time configuration.

## Summary

- Add public `subscribe(topics:)`, `unsubscribe()`, and `subscription()` on `KafkaConsumer`
- `subscribe(topics:)` replaces the current subscription, matching librdkafka semantics
- Add `unsubscribe()` and `subscription()` to `RDKafkaClient`
- Fix force-unwrap crash when `consumptionStrategy` is nil
- Consumers can be created without `consumptionStrategy` and subscribe dynamically

## Test plan
- [x] Unit tests: closed consumer, empty topics, subscribe before run, nil strategy, duplicate topics, regex patterns
- [x] Integration tests: subscription replacement, unsubscribe, re-subscribe, dynamic multi-topic consume
- [x] `swift test --skip IntegrationTests` passes

### DIFF
--- a/Sources/Kafka/KafkaConsumer.swift
+++ b/Sources/Kafka/KafkaConsumer.swift
@@ -388,7 +388,7 @@ public final class KafkaConsumer: Sendable, Service {
     ///
     /// - Returns: An array of topic names or patterns the consumer is currently subscribed to.
     /// - Throws: A ``KafkaError`` if the consumer is closed or the query failed.
-    public func subscription() throws -> [String] {
+    public func subscribedTopics() throws -> [String] {
         let action = self.stateMachine.withLockedValue { $0.withClientForSubscription() }
         switch action {
         case .client(let client):
--- a/Sources/Kafka/KafkaConsumer.swift
+++ b/Sources/Kafka/KafkaConsumer.swift
@@ -334,11 +334,73 @@ public final class KafkaConsumer: Sendable, Service {
         )
     }
 
-    /// Subscribe to the given list of `topics`.
-    /// The partition assignment happens automatically using `KafkaConsumer`'s consumer group.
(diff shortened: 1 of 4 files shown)


Sources/Kafka/KafkaConsumer.swift
```diff
@@ -334,11 +334,73 @@ public final class KafkaConsumer: Sendable, Service {
...
+        case .throwClosedError:
+            throw KafkaError.connectionClosed(reason: "Consumer is closed")
+        }
+    }
+
+    /// Get the current topic subscription.
+    ///
+    /// - Returns: An array of topic names or patterns the consumer is currently subscribed to.
+    /// - Throws: A ``KafkaError`` if the consumer is closed or the query failed.
+    public func subscription() throws -> [String] {
```
FranzBusch: What if we name this `getCurrentlySubcribedTopics`? To convey with the name what this is doing.

ssikka100: Changed to subscribedTopics() to keep the name concise while still conveying intent. It returns the current set of subscriptions, so ‘currently’ felt redundant.

FranzBusch (approved):

PR #228, opened 2026-04-21
Add pause/resume partition APIs and message timestamps

ssikka100:
## Motivation
Pause/resume partition APIs are needed for per-partition flow control. When downstream processing is slow, pausing specific partitions prevents message buildup without stopping the entire consumer. The consumer remains in the group and continues heartbeating while paused.

`KafkaConsumerMessage` has no timestamp field. Timestamps are fundamental to Kafka, they drive time-based windowing, event ordering, and log compaction retention. 

## Summary
- Add `pause(topicPartitions:)` and `resume(topicPartitions:)` on `KafkaConsumer`
- Add `pausePartitions` and `resumePartitions` to `RDKafkaClient`
- Add `timestamp: Int64?` (milliseconds since epoch) and `timestampType: KafkaTimestampType` to `KafkaConsumerMessage` via `rd_kafka_message_timestamp`
- Add `KafkaTimestampType` struct with `.createTime`, `.logAppendTime`, `.notAvailable`

## Test plan
- [x] Unit tests: timestamp type constants/equality/description, pause/resume on closed consumer
- [x] Integration tests: consumed message has timestamp with valid type, pause/resume lifecycle
- [x] `swift test --skip IntegrationTests` passes

### DIFF
--- a/Sources/Kafka/KafkaConsumerMessage.swift
+++ b/Sources/Kafka/KafkaConsumerMessage.swift
@@ -15,6 +15,38 @@
 import Crdkafka
 import NIOCore
 
+/// The type of timestamp on a Kafka message.
+public struct KafkaTimestampType: Hashable, Sendable, CustomStringConvertible {
+    /// The raw value corresponding to the librdkafka timestamp type.
+    public let rawValue: Int32
+
+    public init(rawValue: Int32) {
+        self.rawValue = rawValue
+    }
+
+    /// Timestamp not available.
+    public static let notAvailable = KafkaTimestampType(
+        rawValue: Int32(RD_KAFKA_TIMESTAMP_NOT_AVAILABLE.rawValue)
+    )
+    /// Timestamp set by the producer (message creation time).
+    public static let createTime = KafkaTimestampType(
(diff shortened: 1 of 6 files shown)


Sources/Kafka/KafkaConsumerMessage.swift
```diff
@@ -29,6 +61,10 @@ public struct KafkaConsumerMessage {
     public var value: ByteBuffer
     /// The offset of the message in its partition.
     public var offset: KafkaOffset
+    /// The timestamp of the message in milliseconds since epoch, or `nil` if not available.
+    public var timestamp: Int64?
```
FranzBusch, line 65: Should we use `Date` for this?

ssikka100: I think we should use timestamp since it aligns with the Kafka ecosystem where timestamps are consistently represented as epoch milliseconds. And also librdkafka returns timestamp so it avoids any conversion overhead on every consumed message and keeps precision lossless.

FranzBusch (approved):

PR #239, opened 2026-05-13
refactor: refactor KafkaContiguousBytes protocol

gulivero1773:
Change KafkaContiguousBytes to func withUnsafeBytes<R>(_ body: (UnsafeBufferPointer<UInt8>) throws -> R) rethrows -> R
### Motivation:
When use  KafkaProducerMessage(topic: topic, value: data) then project have error - Instance method 'send' requires that 'Data' conform to 'KafkaContiguousBytes', but I use Data, and Data have KafkaContiguousBytes
### Modifications:
Change KafkaContiguousBytes protocol and change all func for this protocol
### Result:
All extensions for KafkaContiguousBytes changed to new func

### DIFF
--- a/Sources/Kafka/Data/Array+KafkaContiguousBytes.swift
+++ b/Sources/Kafka/Data/Array+KafkaContiguousBytes.swift
@@ -12,4 +12,10 @@
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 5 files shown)


FranzBusch (commented):
We should really remove this protocol and start using the various span types instead. @ssikka100 is this something you can help with?

PR #243, opened 2026-06-22
Add structured logging with metadata enrichment

ssikka100:
## Summary

- Enrich logger with structured metadata (`kafka.client.id`, `kafka.group.id`, `kafka.client.type`) for log filtering
- Add lifecycle and operational logging at appropriate levels (debug/trace/warning/info)
- Extract metadata key strings into `KafkaLoggingKeys` enum with `kafka.*` dot-separated naming
- Fix `producerLog` test to tolerate additional log events

## Motivation

The Swift client bridges librdkafka's internal logs but had no visibility into Swift-layer operations. This adds application-boundary logging that librdkafka cannot provide: lifecycle state, subscription changes, produce failures, and shutdown progress.

## Test plan

- [x] Build passes with `-Xswiftc -warnings-as-errors`
- [x] Unit tests pass

### DIFF
--- a/Sources/Kafka/KafkaConsumer.swift
+++ b/Sources/Kafka/KafkaConsumer.swift
@@ -206,7 +206,15 @@ public final class KafkaConsumer: Sendable, Service {
     ) throws {
         self.config = config
         self.stateMachine = stateMachine
-        self.logger = logger
+        var enrichedLogger = logger
+        if let clientId = config.clientId {
+            enrichedLogger[metadataKey: KafkaLoggingKeys.clientId] = "\(clientId)"
+        }
+        if let groupId = config.groupId {
+            enrichedLogger[metadataKey: KafkaLoggingKeys.groupId] = "\(groupId)"
+        }
+        enrichedLogger[metadataKey: KafkaLoggingKeys.clientType] = "consumer"
+        self.logger = enrichedLogger
         self.eventsSource = eventsSource
         self.rebalanceContext = rebalanceContext
 
@@ -221,6 +229,13 @@ public final class KafkaConsumer: Sendable, Service {
                 rebalanceContext: rebalanceContext
             )
         }
+
+        self.logger.debug(
+            "Kafka consumer initialized",
+            metadata: [
+                KafkaLoggingKeys.bootstrapServers: "\(config.bootstrapServers?.joined(separator: ",") ?? "default")"
+            ]
+        )
     }
 
     /// Creates a new consumer.
@@ -376,6 +391,8 @@ public final class KafkaConsumer: Sendable, Service {
             return
         }
 
+        self.logger.debug("Subscribing to topics", metadata: [KafkaLoggingKeys.topics: "\(topics)"])
+
         let action = self.stateMachine.withLockedValue { $0.withClientForSubscription() }
         switch action {
         case .client(let client):
@@ -403,6 +420,8 @@ public final class KafkaConsumer: Sendable, Service {
         let action = self.stateMachine.withLockedValue { $0.withClientForSubscription() }
         switch action {
         case .client(let client):
+            let currentTopics = (try? client.subscription()) ?? []
+            self.logger.debug("Unsubscribing from all topics", metadata: [KafkaLoggingKeys.topics: "\(currentTopics)"])
             try client.unsubscribe()
         case .throwClosedError:
             throw KafkaError.connectionClosed(reason: "Consumer is closed")
@@ -554,7 +573,7 @@ public final class KafkaConsumer: Sendable, Service {
                 do {
                     try client.consumerClose()
                 } catch {
-                    self.logger.error("Closing KafkaConsumer failed", metadata: ["error": "\(error)"])
+                    self.logger.info("Closing KafkaConsumer failed", error: error)
                     self.stateMachine.withLockedValue { $0.closeFailed() }
                 }
                 self.pollAndDrainEvents(client: client)
@@ -571,6 +590,7 @@ public final class KafkaConsumer: Sendable, Service {
                 if let client {
                     self.finalDrain(client: client)
                 }
+                self.logger.debug("Kafka consumer shut down")
                 return
             }
         }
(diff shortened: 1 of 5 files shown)


Sources/Kafka/KafkaConsumer.swift
```diff
@@ -206,7 +206,15 @@ public final class KafkaConsumer: Sendable, Service {
     ) throws {
         self.config = config
         self.stateMachine = stateMachine
-        self.logger = logger
+        var enrichedLogger = logger
+        if let clientId = config.clientId {
+            enrichedLogger[metadataKey: "kafka.client_id"] = "\(clientId)"
```
FranzBusch: Can we extract the metadata key string values into static var's? Something like this

```
enum LoggingKeys {
    static var clientID = "kafka.client.id"
}
```

We also prefer dot separation instead of underscore

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -221,6 +229,13 @@ public final class KafkaConsumer: Sendable, Service {
                 rebalanceContext: rebalanceContext
             )
         }
+
+        self.logger.info(
```
FranzBusch: Libraries should not log at `info` level unless they discovered an error. I would recommend putting this at `debug` level.

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -221,6 +229,13 @@ public final class KafkaConsumer: Sendable, Service {
...
+            "Kafka consumer initialized",
+            metadata: [
+                "bootstrap_servers": "\(config.bootstrapServers?.joined(separator: ",") ?? "default")",
```
FranzBusch: Same comment here for extracting the metadata key. Can we keep the `kafka.` prefixing for all the keys please

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -376,6 +391,8 @@ public final class KafkaConsumer: Sendable, Service {
             return
         }
 
+        self.logger.info("Subscribing to topics", metadata: ["topics": "\(topics)"])
```
FranzBusch: Same comment here and for the rest of the diff for metadata keys and info level

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -934,11 +954,10 @@ public final class KafkaConsumer: Sendable, Service {
         do {
             try client.consumerClose()
         } catch {
-            if let error = error as? KafkaError {
-                logger.error("Closing KafkaConsumer failed: \(error.description)")
-            } else {
-                logger.error("Caught unknown error: \(error)")
-            }
+            logger.error(
```
FranzBusch: This should be logged at `info` level

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -400,6 +417,7 @@ public final class KafkaConsumer: Sendable, Service {
     ///
     /// - Throws: A ``KafkaError`` if the consumer is closed or unsubscribing failed.
     public func unsubscribe() throws {
+        self.logger.info("Unsubscribing from all topics")
```
FranzBusch: Would be good to include all currently subscribed topics if possible

Sources/Kafka/KafkaProducer.swift
```diff
@@ -303,10 +315,21 @@ public final class KafkaProducer: Service, Sendable {
                     source?.finish()
                     self.stateMachine.withLockedValue { $0.failPendingContinuations() }
                 }
-                try await client.flush(timeoutMilliseconds: Int32(self.config.shutdownFlushTimeoutMs))
+                self.logger.info("Flushing outstanding messages before shutdown")
```
FranzBusch: This should be at trace level

Sources/Kafka/KafkaProducer.swift
```diff
@@ -303,10 +315,21 @@ public final class KafkaProducer: Service, Sendable {
...
+                do {
+                    try await client.flush(timeoutMilliseconds: Int32(self.config.shutdownFlushTimeoutMs))
+                } catch {
+                    self.logger.warning(
```
FranzBusch: This should be at info level

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -554,7 +573,7 @@ public final class KafkaConsumer: Sendable, Service {
                 do {
                     try client.consumerClose()
                 } catch {
-                    self.logger.error("Closing KafkaConsumer failed", metadata: ["error": "\(error)"])
+                    self.logger.info("Closing KafkaConsumer failed", metadata: [KafkaLoggingKeys.error: "\(error)"])
```
FranzBusch: Instead of passing error as metadata you can pass the error to the method directly. swift-log added support for that in recent versions. Applies to all the places where you pass errors

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -589,7 +609,7 @@ public final class KafkaConsumer: Sendable, Service {
                 }
                 self.logger.error(
```
FranzBusch: This should become info level and pass the error directly

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -618,8 +638,8 @@ public final class KafkaConsumer: Sendable, Service {
             self.logger.info(
```
FranzBusch: This should be debug

Sources/Kafka/KafkaConsumer.swift
```diff
@@ -644,7 +664,7 @@ public final class KafkaConsumer: Sendable, Service {
                 }
                 self.logger.error(
```
FranzBusch: this should be info

Sources/Kafka/KafkaProducer.swift
```diff
@@ -361,11 +401,19 @@ public final class KafkaProducer: Service, Sendable {
...
-                topicHandles: topicHandles
-            )
+            do {
+                try client.produce(
+                    message: message,
+                    newMessageID: newMessageID,
+                    topicHandles: topicHandles
+                )
+            } catch {
+                self.logger.warning(
```
FranzBusch: this should be info

Sources/Kafka/KafkaConsumer.swift
ssikka100: I see the error: parameter support requires swift-log versions that use swift-tools-version 6.2+. Should we drop Swift 6.0/6.1 support, or keep the metadata approach for now?

FranzBusch (approved):

PR #247, opened 2026-07-10
Regenerate configuration files from librdkafka 2.14.1

ssikka100:
### Motivation

[PR #236](https://github.com/swift-server/swift-kafka-client/pull/236) bumped librdkafka to 2.14.1 but never re-ran `make gyb`, so the generated Swift configuration files fell out of sync with `Sources/Crdkafka/librdkafka/CONFIGURATION.md`.

Users can't set librdkafka 2.14.1's newly introduced options from Swift because the wrapper properties don't exist. Running `make gyb` today would expose the drift.

### Changes

Regenerate the three generated configuration files with `python3.13 dev/gyb.py`. No hand-written changes; the `.gyb` templates are unmodified.

* `KafkaConfig.swift` — four new nested option types:
  * `MetadataRecoveryStrategy` (`.none`, `.rebootstrap`)
  * `SaslOauthbearerGrantType` (`.client_credentials`, `.urn:ietf:params:oauth:grant-type:jwt-bearer`)
  * `SaslOauthbearerAssertionAlgorithm` (`.RS256`, `.ES256`)
  * `SaslOauthbearerMetadataAuthenticationType` (`.none`, `.azure_imds`)
* `KafkaConsumerConfig.swift` — new `var` properties surfacing the matching librdkafka 2.14.1 consumer options.
* `KafkaProducerConfig.swift` — new `var` properties surfacing the matching librdkafka 2.14.1 producer options.


ssikka100:
Closing to re-scope this change with a tighter diff. Will resubmit.

PR #250, opened 2026-07-23
1.0 Beta release API changes

ssikka100:
### Consumer

- Single construction path: `makeConsumer(config:logger:) -> (consumer:, events:)`. Removed the plain `init` and `makeConsumerWithEvents`.
- Removed `scheduleCommit(_:)` / `scheduleCommit()` — silently dropped errors. Use `Task { try? await consumer.commit(msg) }` instead.
- Renamed `commit()` → `commitStoredOffsets()`. `commit(_:)` unchanged.
- Removed deprecated pass-throughs: `init(configuration:)`, `makeConsumerWithEvents(configuration:)`, `commitSync(_:)`.

### Producer

- Single construction path: `makeProducer(config:logger:) -> (producer:, events:)`. Removed the plain `init` and `makeProducerWithEvents`.
- Fixed duplicate delivery reports: `sendAndAwait(_:)` reports now resolve only the awaiting call, no longer also emitted on `KafkaProducerEvents`.
- Removed deprecated pass-throughs: `init(configuration:)`, `makeProducerWithEvents(configuration:)`.

PR #252, opened 2026-07-28
API review: consumer and producer APIs [Not to be merged]

ssikka100:
## API review — consumer and producer

This PR is for API Review only and not to be merged. 

Documentation companion to #250. Not a code change — this is the design-decision record for the consumer and producer API surface, for review alongside the implementation PR.

### What's here

- `Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md` — `KafkaConsumer` proposed API, with resolved design decisions (offset-store naming, fire-and-forget commit removal, single construction path) and raligned with other language clients(rust/python/java).
- `Sources/Kafka/Kafka.docc/APIReview/0002-producer-api.md` — `KafkaProducer` proposed API, with resolved design decisions (delivery-report correlation, duplicate-report fix, single construction path) and aligned with other language clients.

### DIFF
--- a/Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
+++ b/Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
@@ -1,0 +1,355 @@
+# Swift-kafka-client-0001: Consumer API
+
+## Overview
+
+- **Proposal:** Swift-kafka-client-0001
+- **Implementation PR:** [swift-server/swift-kafka-client#250](https://github.com/swift-server/swift-kafka-client/pull/250) — tracks this proposal's implementation
+- **Note:** This PR exists to collect API review feedback. Once the design is finalized and landed in #250, this PR will be closed rather than merged.
+
+## Introduction
+
+`KafkaConsumer` is the entry point for reading records from a Kafka cluster. It integrates with `swift-service-lifecycle` (`Service`) and exposes records through an `AsyncSequence`.
+
+## Motivation
+
+Kafka's consumer semantics cover four distinct axes:
+
+- Offset persistence — automatic (auto-store + auto-commit), manual per-message, or batched.
+- Subscription vs assignment — dynamic group membership vs pinned partitions.
+- Backpressure — pause/resume without leaving the group.
+- Rebalance observation — react to partition assignment changes.
+
+This proposal freezes the 1.0 Beta shape of the consumer surface.
+
+## Proposed API
+
+```swift
+/// Consumes messages from a Kafka cluster as part of a service lifecycle.
+public final class KafkaConsumer: Sendable, Service {
+
+    // MARK: Creation
+
+    /// Creates a consumer and its paired message and events sequences.
+    ///
+    /// Every consumer receives an events sequence for rebalance and error observation.
+    /// Callers who don't need it discard the returned `events` value.
+    ///
+    /// - Parameter config: The ``KafkaConsumerConfig`` for configuring the consumer.
+    /// - Returns: A named tuple containing the created ``KafkaConsumer`` and its ``KafkaConsumer/Messages`` and ``KafkaConsumer/Events`` `AsyncSequence`s.
+    /// - Throws: A ``KafkaError`` if the initialization failed.
+    public static func makeConsumer(
+        config: KafkaConsumerConfig
+    ) throws -> (consumer: KafkaConsumer, messages: Messages, events: Events)
+
+    /// Creates a consumer, runs it for the duration of the closure, and shuts it down on return.
+    ///
+    /// Spawns the consumer's `run()` loop internally and cancels it when `body` returns or
+    /// throws — no `ServiceGroup` required. Intended for scoped, short-lived use (tests,
+    /// scripts, one-off jobs). For long-running services, use `makeConsumer(config:)`
+    /// with a `ServiceGroup`.
+    ///
+    /// - Parameters:
+    ///   - config: The ``KafkaConsumerConfig`` for configuring the consumer.
+    ///   - body: A closure that receives the consumer and its message and events sequences.
+    /// - Returns: The value returned by `body`.
+    /// - Throws: A ``KafkaError`` if initialization failed, or any error thrown by `body`.
+    public static func withConsumer<Result>(
+        config: KafkaConsumerConfig,
+        _ body: (KafkaConsumer, Messages, Events) async throws -> Result
+    ) async throws -> Result
+
+    // MARK: Service conformance
+
+    /// Starts the consumer's event loop.
+    ///
+    /// This method must be called for the consumer to start receiving records and events.
+    /// It returns once the consumer is closed or the task is cancelled.
+    public func run() async throws
+
+    // MARK: Subscription management
+
+    /// Subscribes to the given topics.
+    ///
+    /// - Parameter topics: The list of topics to subscribe to.
+    /// - Throws: A ``KafkaError`` if subscribing failed or the consumer is closed.
+    public func subscribe(topics: [KafkaTopic]) throws
+
+    /// Unsubscribes from all currently subscribed topics.
+    ///
+    /// - Throws: A ``KafkaError`` if unsubscribing failed or the consumer is closed.
+    public func unsubscribe() throws
+
+    /// The topics the consumer is currently subscribed to.
+    public var subscribedTopics: [KafkaTopic] { get throws }
+
+    // MARK: Pause / resume
+
+    /// Pauses consumption for the given topic-partitions.
+    ///
+    /// - Parameter topicPartitions: The topic-partitions to pause.
+    /// - Throws: A ``KafkaError`` if pausing failed or the consumer is closed.
+    public func pause(topicPartitions: [KafkaTopicPartition]) throws
+
+    /// Resumes consumption for the given topic-partitions.
+    ///
+    /// - Parameter topicPartitions: The topic-partitions to resume.
+    /// - Throws: A ``KafkaError`` if resuming failed or the consumer is closed.
+    public func resume(topicPartitions: [KafkaTopicPartition]) throws
+
+    // MARK: Offset management
+
+    /// Records an offset in librdkafka's internal store for later commit.
+    ///
+    /// Requires `enableAutoOffsetStore == false`.
+    ///
+    /// - Parameter message: The message whose offset should be stored.
+    /// - Throws: A ``KafkaError`` if storing the offset failed or the consumer is closed.
+    public func storeOffset(_ message: Message) throws
+
+    /// Commits the offset of one message directly. Bypasses the offset store.
+    ///
+    /// - Parameter message: The message whose offset should be committed.
+    /// - Throws: A ``KafkaError`` if committing failed or the consumer is closed.
+    public func commit(_ message: Message) async throws
+
+    /// Commits every offset currently in the local offset store.
+    ///
+    /// - Throws: A ``KafkaError`` if committing failed or the consumer is closed.
+    public func commitStoredOffsets() async throws
+
+    // MARK: Position query
+
+    /// Retrieves the last-committed offsets from the broker for the given topic-partitions.
+    ///
+    /// - Parameters:
+    ///   - topicPartitions: The topic-partitions to query.
+    ///   - timeout: The maximum time to wait for the broker to respond.
+    /// - Returns: A list of topic-partition offsets.
+    /// - Throws: A ``KafkaError`` if querying failed or the consumer is closed.
+    public func committed(
+        topicPartitions: [KafkaTopicPartition],
+        timeout: Duration
+    ) async throws -> [KafkaTopicPartitionOffset]
+
+    /// Retrieves the current positions (next offset to be fetched) for the given topic-partitions.
+    ///
+    /// - Parameter topicPartitions: The topic-partitions to query.
+    /// - Returns: A list of topic-partition offsets.
+    /// - Throws: A ``KafkaError`` if querying failed or the consumer is closed.
+    public func position(
+        topicPartitions: [KafkaTopicPartition]
+    ) throws -> [KafkaTopicPartitionOffset]
+
+    /// A boolean indicating whether the consumer's assignment has been lost.
+    public var isAssignmentLost: Bool { get }
+
+    // MARK: Seeking
+
+    /// Seeks to a specific offset for each of the given topic-partitions.
+    ///
+    /// - Parameters:
+    ///   - topicPartitionOffsets: The offsets to seek to.
+    ///   - timeout: The maximum time to wait for the seek to complete.
+    /// - Throws: A ``KafkaError`` if seeking failed or the consumer is closed.
+    public func seek(
+        topicPartitionOffsets: [KafkaTopicPartitionOffset],
+        timeout: Duration
+    ) async throws
+}
+```
+
+### Message and event sequences
+
+```swift
+extension KafkaConsumer {
+    /// An `AsyncSequence` of records consumed from the Kafka cluster.
+    public struct Messages: Sendable, AsyncSequence {
+        public typealias Element = Message
+        public struct AsyncIterator: AsyncIteratorProtocol {
+            public mutating func next() async throws -> Element?
+        }
+        public func makeAsyncIterator() -> AsyncIterator
+    }
+
+    /// An `AsyncSequence` of rebalance and error events for the consumer.
+    public struct Events: Sendable, AsyncSequence {
+        public typealias Element = Event
+        public struct AsyncIterator: AsyncIteratorProtocol {
+            public mutating func next() async -> Element?
+        }
+        public func makeAsyncIterator() -> AsyncIterator
+    }
+
+    /// An event observed on the consumer's events sequence.
+    @nonexhaustive
+    public enum Event: Sendable, Hashable {
+        /// The consumer's partition assignment changed.
+        case rebalance(Rebalance)
+        /// A non-fatal error occurred.
+        case error(KafkaError)
+    }
+
+    /// A change in the consumer's partition assignment.
+    public struct Rebalance: Sendable, Hashable {
+        /// The kind of rebalance that occurred.
+        public enum Kind: Sendable, Hashable {
+            /// Partitions were assigned to this consumer.
+            case assign
+            /// Partitions were revoked from this consumer.
+            case revoke
+            /// The rebalance failed with the given reason.
+            case error(String)
+        }
+        /// Whether partitions were assigned, revoked, or failed to rebalance.
+        public var kind: Kind { get }
+        /// The topic-partitions affected by this rebalance.
+        public var partitions: [KafkaTopicPartition] { get }
+    }
+
+    /// A single record consumed from a Kafka cluster.
+    public struct Message: Sendable {
+        /// The topic this message was consumed from.
+        public var topic: KafkaTopic { get }
+        /// The partition this message was consumed from.
+        public var partition: KafkaPartition { get }
+        /// This message's offset within its partition.
+        public var offset: KafkaOffset { get }
+        /// The message key, if any.
+        public var key: ByteBuffer? { get }
+        /// The message value.
+        public var value: ByteBuffer { get }
+        /// The message headers.
+        public var headers: [KafkaHeader] { get }
+        /// The timestamp of the message in milliseconds since epoch, or `nil` if not available.
+        public var timestamp: Int64? { get }
+        /// The type of timestamp on this message.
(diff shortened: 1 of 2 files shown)


Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+    // MARK: Creation
+
+    /// Creates a consumer and its paired events sequence.
+    ///
+    /// Every consumer receives an events sequence for rebalance and error observation.
+    /// Callers who don't need it discard the returned `events` value.
+    public static func makeConsumer(
+        configuration: KafkaConsumerConfiguration,
+        logger: Logger
```
czechboy0: I don't think we need to ask users to explicitly pass Logger anymore - the initializer can either grab Logger.current automatically, or it can use Logger.current in methods where it needs to log (which will include metadata of the task that invoked each method).

Details: https://swiftpackageindex.com/apple/swift-log/1.14.0/documentation/logging/slg-0006-task-local-logger

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,172 @@
...
+    ) throws -> (consumer: KafkaConsumer, events: KafkaConsumerEvents)
+
+    // MARK: Service conformance
+
+    public func run() async throws
+    public func triggerGracefulShutdown()
```
czechboy0: Why do we need the explicit `triggerGracefulShutdown`? The consumer could remove this and instead use Swift Service Lifecycle's graceful shutdown detection.

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,172 @@
...
+        topicPartitionOffsets: [KafkaTopicPartitionOffset],
+        timeout: Duration
+    ) async throws
+}
+```
+
+### Message and event sequences
+
+```swift
+public struct KafkaConsumerMessages: Sendable, AsyncSequence {
```
czechboy0: It'd be easier to review the public API if they all contained some doc comments.

ssikka100: done

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,253 @@
...
+    /// Starts the consumer's event loop.
+    ///
+    /// This method must be called for the consumer to start receiving records and events.
+    /// It returns once the consumer is closed or the task is cancelled.
+    public func run() async throws
+
+    // MARK: Message consumption
+
+    /// An `AsyncSequence` of records consumed from the Kafka cluster.
+    public var messages: KafkaConsumerMessages { get }
```
FranzBusch: I think we should return this from the `make` method as well to align with the events sequence

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,253 @@
...
+
+    /// Creates a consumer and its paired events sequence.
+    ///
+    /// Every consumer receives an events sequence for rebalance and error observation.
+    /// Callers who don't need it discard the returned `events` value.
+    ///
+    /// - Parameter configuration: The ``KafkaConsumerConfiguration`` for configuring the consumer.
+    /// - Returns: A named tuple containing the created ``KafkaConsumer`` and its ``KafkaConsumerEvents`` `AsyncSequence`.
+    /// - Throws: A ``KafkaError`` if the initialization failed.
+    public static func makeConsumer(
```
FranzBusch, line 40: In addition to this we should also provide a `static func withConsumer` method that provides the consumer and the sequences to a closure

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,253 @@
...
+    /// Unsubscribes from all currently subscribed topics.
+    ///
+    /// - Throws: A ``KafkaError`` if unsubscribing failed or the consumer is closed.
+    public func unsubscribe() throws
+
+    /// Returns the topics the consumer is currently subscribed to.
+    ///
+    /// - Returns: A list of subscribed topic names.
+    /// - Throws: A ``KafkaError`` if the consumer is closed.
+    public func subscribedTopics() throws -> [String]
```
FranzBusch: Should be a computed property instead of a method.

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,253 @@
...
+
+    /// Seeks to a specific offset for each of the given topic-partitions.
+    ///
+    /// - Parameters:
+    ///   - topicPartitionOffsets: The offsets to seek to.
+    ///   - timeout: The maximum time to wait for the seek to complete.
+    /// - Throws: A ``KafkaError`` if seeking failed or the consumer is closed.
+    public func seek(
+        topicPartitionOffsets: [KafkaTopicPartitionOffset],
+        timeout: Duration
```
FranzBusch, line 154: Remind me again why this method needs a timeout and the others don’t

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,253 @@
...
+public struct KafkaConsumerMessages: Sendable, AsyncSequence {
+    public typealias Element = KafkaConsumerMessage
+    public struct AsyncIterator: AsyncIteratorProtocol {
+        public mutating func next() async throws -> Element?
+    }
+    public func makeAsyncIterator() -> AsyncIterator
+}
+
+/// An `AsyncSequence` of rebalance and error events for the consumer.
+public struct KafkaConsumerEvents: Sendable, AsyncSequence {
```
FranzBusch: Let’s nest this type in the consumer class

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,253 @@
...
+    public typealias Element = KafkaConsumerEvent
+    public struct AsyncIterator: AsyncIteratorProtocol {
+        public mutating func next() async -> Element?
+    }
+    public func makeAsyncIterator() -> AsyncIterator
+}
+
+/// An event observed on the consumer's events sequence.
+public enum KafkaConsumerEvent: Sendable, Hashable {
```
FranzBusch: Same here. Let’s nest this type

FranzBusch: Let’s make this enum `@nonexhaustive`

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,253 @@
...
+    /// The consumer's partition assignment changed.
+    case rebalance(KafkaConsumerRebalance)
+    /// A non-fatal error occurred.
+    case error(KafkaError)
+}
+
+/// A change in the consumer's partition assignment.
+public struct KafkaConsumerRebalance: Sendable, Hashable {
```
FranzBusch: Same here let’s nest this

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,253 @@
...
+    /// The kind of rebalance that occurred.
+    public enum Kind: Sendable, Hashable {
```
FranzBusch: Do we think there ever can be another event kind in the future?

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,253 @@
...
+    /// Whether partitions were assigned, revoked, or failed to rebalance.
+    public var kind: Kind { get }
+    /// The topic-partitions affected by this rebalance.
+    public var partitions: [KafkaTopicPartition] { get }
+}
+
+/// A single record consumed from a Kafka cluster.
+public struct KafkaConsumerMessage: Sendable {
+    /// The topic this message was consumed from.
+    public var topic: String { get }
```
FranzBusch: I think it’s worth introducing a concrete type for topics that just wrap a string

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,253 @@
...
+    /// This message's offset within its partition.
+    public var offset: KafkaOffset { get }
+    /// The message key, if any.
+    public var key: [UInt8]? { get }
+    /// The message value.
+    public var value: [UInt8] { get }
+    /// The message headers.
+    public var headers: [KafkaHeader] { get }
+    /// The message's broker-assigned timestamp, if any.
+    public var timestamp: KafkaTimestampType? { get }
```
FranzBusch: Can we include this in the API review here

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,253 @@
...
+    public func commit() async throws                                           // parameterless overload
+    public func scheduleCommit(_ message: KafkaConsumerMessage) throws          // fire-and-forget per message
+    public func scheduleCommit() throws                                         // fire-and-forget bulk
+
+    // ...subscribe/unsubscribe/subscribedTopics/
+    //    pause/resume/committed/position/isAssignmentLost/seek unchanged
+}
+```
+
+### What has changed
```
FranzBusch: Can we remove this diff section

FranzBusch (commented):
Review for the consumer API

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
ssikka100: Are you suggesting spawning run() in a task group internally, hand the closure (consumer, messages, events) and cancel the group once it returns and no ServiceGroup. ? 
Is this aimed at cases that don't want the ServiceGroup such as tests, scripts rather than being an alternative for production services?

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
ssikka100: seek() and committed() both wrap blocking librdkafka calls (rd_kafka_seek_partitions, rd_kafka_committed) that take a timeout_ms argument and get offloaded to a background queue so they don't block the event loop. 
On the other hand, position() is a local cache read, no broker round-trip, so there's nothing to time out. Same reasoning for pause/resume/subscribe/unsubscribe as well they don't wait on the broker either.

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
ssikka100: I don't think there is another outcome possible in future since librdkafka only exposes these two rebalance outcomes (and error) only.

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
ssikka100: I agree, this makes more sense and great call. Made the changes accordingly

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
ssikka100: @nonexhaustive requires 6.2.3+ however right now the package floor is 6.1. Should we drop support for 6.1 ?

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
FranzBusch: Yes correct.

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
FranzBusch: Is the timeout passed through to `rdkafka`? Is there a way to manually cancel the seek or commit before the timeout i.e. what happens if the task got cancelled that is waiting for either of those operations?

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
FranzBusch: Yeah we should move to 6.2.3. since that allows us to use `@nonexhaustive`

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
ssikka100: yes the timeout is passed straight to librdkafka methods. There 's no way to manually cancel these in-flight calls because librdkafka doesn't expose any cancellation handle. 

In current implementation, if the awaiting task is cancelled nothing is interrupted, it stays suspended until the call returns or the timeout hits and the continuation then resumes the same caller with the normal result (not CancellationError) — Task.isCancelled is just set. 
We could add a withTaskCancellationHandler to return CancellationError early but the librdkafka call would still run to completion (result/timeout) and that result would simply be dropped.

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,355 @@
...
+    public static func makeConsumer(
+        config: KafkaConsumerConfig
+    ) throws -> (consumer: KafkaConsumer, messages: Messages, events: Events)
+
+    /// Creates a consumer, runs it for the duration of the closure, and shuts it down on return.
+    ///
+    /// Spawns the consumer's `run()` loop internally and cancels it when `body` returns or
+    /// throws — no `ServiceGroup` required. Intended for scoped, short-lived use (tests,
+    /// scripts, one-off jobs). For long-running services, use `makeConsumer(config:)`
+    /// with a `ServiceGroup`.
```
FranzBusch:
```suggestion
    /// Spawns the consumer's `run()` loop internally and cancels it when `body` returns or
    /// throws.
```

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,355 @@
...
+        /// The topic this message was consumed from.
+        public var topic: KafkaTopic { get }
+        /// The partition this message was consumed from.
+        public var partition: KafkaPartition { get }
+        /// This message's offset within its partition.
+        public var offset: KafkaOffset { get }
+        /// The message key, if any.
+        public var key: ByteBuffer? { get }
+        /// The message value.
+        public var value: ByteBuffer { get }
```
FranzBusch: We should use `[UInt8]` here instead of `ByteBuffer`

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,355 @@
...
+Wraps the underlying librdkafka timestamp-type value, exposing named constants instead of raw integers — same pattern as ``KafkaPartition``/``KafkaOffset``.
+
+```swift
+/// The type of timestamp on a Kafka message.
+public struct KafkaTimestampType: Hashable, Sendable, CustomStringConvertible {
+    /// The raw value corresponding to the librdkafka timestamp type.
+    public let rawValue: Int32
+
+    /// Creates a timestamp type from its raw librdkafka value.
+    public init(rawValue: Int32)
```
FranzBusch: I don't think we should expose this publicly

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,355 @@
...
+
+Shared with the producer proposal. Each wraps an underlying primitive — matching the ``KafkaPartition``/``KafkaOffset`` pattern — so call sites use named types instead of bare `String`/`Int` values.
+
+```swift
+/// A header attached to a Kafka message.
+public struct KafkaHeader: Sendable, Hashable {
+    /// The key associated with the header.
+    public var key: String
+    /// The value associated with the header.
+    public var value: ByteBuffer?
```
FranzBusch: Same here let's use an array instead

Sources/Kafka/Kafka.docc/APIReview/0001-consumer-api.md
```diff
@@ -0,0 +1,355 @@
...
+    /// throws — no `ServiceGroup` required. Intended for scoped, short-lived use (tests,
+    /// scripts, one-off jobs). For long-running services, use `makeConsumer(config:)`
+    /// with a `ServiceGroup`.
+    ///
+    /// - Parameters:
+    ///   - config: The ``KafkaConsumerConfig`` for configuring the consumer.
+    ///   - body: A closure that receives the consumer and its message and events sequences.
+    /// - Returns: The value returned by `body`.
+    /// - Throws: A ``KafkaError`` if initialization failed, or any error thrown by `body`.
+    public static func withConsumer<Result>(
```
FranzBusch:
```suggestion
    public static func withConsumer<Result: ~Copyable>(
```

Sources/Kafka/Kafka.docc/APIReview/0002-producer-api.md
```diff
@@ -0,0 +1,178 @@
...
+    public static func makeProducer(
+        config: KafkaProducerConfig
+    ) throws -> (producer: KafkaProducer, events: Events)
+
+    /// Creates a producer, runs it for the duration of the closure, and shuts it down on return.
+    ///
+    /// Spawns the producer's `run()` loop internally and cancels it when `body` returns or
+    /// throws — no `ServiceGroup` required. Intended for scoped, short-lived use (tests,
+    /// scripts, one-off jobs). For long-running services, use `makeProducer(config:)`
+    /// with a `ServiceGroup`.
```
FranzBusch:
```suggestion
    /// Spawns the producer's `run()` loop internally and cancels it when `body` returns or
    /// throws.
```

Sources/Kafka/Kafka.docc/APIReview/0002-producer-api.md
```diff
@@ -0,0 +1,178 @@
...
+    ///
+    /// - Parameters:
+    ///   - config: The ``KafkaProducerConfig`` for configuring the producer.
+    ///   - body: A closure that receives the producer and its events sequence.
+    /// - Returns: The value returned by `body`.
+    /// - Throws: A ``KafkaError`` if initialization failed, or any error thrown by `body`.
+    public static func withProducer<Result>(
```
FranzBusch:
```suggestion
    public static func withProducer<Result: ~Copyable>(
```

Sources/Kafka/Kafka.docc/APIReview/0002-producer-api.md
```diff
@@ -0,0 +1,178 @@
...
+
+    /// Enqueues a record for delivery.
+    ///
+    /// Returns immediately with an identifier that will appear on the eventual ``KafkaProducer/DeliveryReport``.
+    /// Outcomes are reported through the ``KafkaProducer/Events`` sequence.
+    ///
+    /// - Parameter message: The message to send.
+    /// - Returns: A unique identifier for the message.
+    /// - Throws: A ``KafkaError`` if enqueuing the message failed or the producer is closed.
+    public func send<Key: KafkaContiguousBytes, Value: KafkaContiguousBytes>(
```
FranzBusch, line 78: We can respell this now. We don't need the `KafkaContinguousBytes` anymore. Instead I think we can use the `Span` types for this now.

ssikka100: @FranzBusch  I think we should migrate to Span post-1.x rather than now. Two reasons: it needs some API design changes (dropping the stored Message, reshaping send) and  the bigger one is that the span-creation accessors (Array.span, String.utf8Span, ByteBuffer.readableBytesSpan, etc.) are @available(macOS 26, *).

FranzBusch: Migrating post 1.0 is gonna be hard because this requires an API break potentially. Also not doing it before the 1.0 means we are shipping protocols that we really don't need i.e. `KafkaContiguousBytes`. What do you think?

PR #255, opened 2026-08-11
Add pull request benchmarks for allocations and instructions

ssikka100:
### Motivation

We had no pull-request performance signal: allocations and instruction counts could regress silently. This adds [package-benchmark](https://github.com/ordo-one/package-benchmark)-based micro-benchmarks that run on each PR, reusing the same benchmark framework `swift-nio` provides.

### What this adds

**Deterministic micro-benchmarks** (`SwiftKafkaMicroBenchmarks`) — broker-free, pure-Swift hot paths measured for per-operation allocations (`mallocCountTotal` + `objectAllocCount`):
- `KafkaProducerMessage` construction — one benchmark per supported payload type, plus one carrying headers
- `KafkaTopicPartition` / `KafkaTopicPartitionOffset` input-list construction
- `KafkaProducerConfig` / `KafkaConsumerConfig` construction

**Mock-cluster benchmark** (`SwiftKafkaMockBenchmarks`) — a `send`+ack round-trip against librdkafka's in-process mock broker. It's an end-to-end benchmark through a real (in-process) broker, so it's noisy by nature — it lives in its own manual-only workflow (`benchmark-mock.yml`, `workflow_dispatch` / `KAFKA_MOCK_BENCHMARK`) and never runs on pull requests.

### How it runs

`.github/workflows/benchmark.yml` reuses NIO's benchmark framework — `generate_matrix.sh` + `swift_test_matrix.yml` + `check_benchmark_thresholds.sh` — with `libsasl2-dev`/`libssl-dev`/`libjemalloc-dev` added to the setup so librdkafka builds. It checks against committed per-Swift-version thresholds in `Benchmarks/Thresholds/<version>/`.

The gate is scoped for stability:
- **Only the deterministic micro-benchmarks are gated** — the mock benchmark and the broker-based consumer/producer benchmarks (gated behind `KAFKA_LIVE_BENCHMARKS`) never run in CI.
- **Runs on 6.2 / 6.3 and nightly-next.** nightly-main is excluded (Swift trunk is too volatile for stable baselines) and 6.1 is dropped.
- **Only deterministic allocation counts** (`mallocCountTotal` + `objectAllocCount`) with a small absolute tolerance (the swift-nio approach). The retain/release (ARC-traffic) counts, throughput, and resident memory are noisy and are not gated.

### Library change

`KafkaConsumerConfig`/`KafkaProducerConfig.additionalConfig` is exposed via `@_spi(Internal)` so the mock benchmark can enable `test.mock.num.brokers`. Additive and SPI-only — no stable public API change (API-breakage check passes).

### Notes

- The first benchmark run has no committed thresholds, so it generates them, prints them as a git diff in the job log, and fails once (NIO's bootstrap). Those JSON files are added to `Benchmarks/Thresholds/<version>/` and committed to the branch; subsequent runs then pass. The benchmark job is advisory, not a required check.
- `Benchmarks/Package.swift` deployment target bumped to macOS 15 to match the `Kafka` library floor.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,54 @@
+name: Benchmarks
+
+permissions:
+  contents: read
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "**.swift"
+      - "**.yml"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # PR gate: the deterministic, broker-free micro-benchmarks, checked against the
+  # committed per-Swift-version absolute thresholds in Benchmarks/Thresholds/. The
+  # noisy mock-cluster produce benchmark is not gated here; it lives in
+  # benchmark-mock.yml on a manual trigger and never runs on pull requests.
+  construct-benchmark-matrix:
+    name: Construct Benchmarks matrix
+    runs-on: ubuntu-latest
+    outputs:
+      benchmark-matrix: "${{ steps.generate-matrix.outputs.benchmark-matrix }}"
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - id: generate-matrix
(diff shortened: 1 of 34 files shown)


.github/workflows/benchmark.yml
FranzBusch, line 1: Any reason why this is not reusing the workflows provided by NIO?

ssikka100: Changed to use NIO

FranzBusch (approved):

